PR #64, opened 2015-03-27
Relaxed rules on allowable urls in Arc files

navtis:
java.net rules for URI parsing enforce restricted characters on querystrings, causing droid to exit in the middle of an ARC file when it met any.


seamang:
Added further improvements to ARC handling

PR #113, opened 2016-09-13
Blank profile on new signature file.

snail1966:
Fixes an issue whereby the first profile run after installing a new signature file (or reverting to an earlier one) is sometimes blank.


snail1966:
Updated in new request 117 where I just added a carriage return.  Unclear why we were getting checkstyle errors in Travis before as build fine locally..

PR #131, opened 2017-02-01
Support invoking DROID outside its home directory

david-russo:
This change updates the Windows and Unix scripts to allow them to be invoked from outside their installation directory or through symbolic links, and guards against paths which include spaces.

Tested on MacOS 10.12, MinGW64, and Windows 10.


rhubner:
Thank you. We shall include it in the next release.

Issue #74, opened 2015-11-02
Saving report as text sometimes fails


gmcgath:
In the GUI version of DROID 6.1.5, if I create a report and then export it, leaving the default format of Text, it goes through the motions of selecting a file, but no file is created. If I change to another format and then back to Text, it works. 

The failure to output is accompanied by a stack dump; here's the beginning of it.

Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException
    at uk.gov.nationalarchives.droid.gui.report.ExportReportAction.execute(ExportReportAction.java:86)
    at uk.gov.nationalarchives.droid.gui.report.ReportViewFrame.exportButtonActionPerformed(ReportViewFrame.java:244)
    at uk.gov.nationalarchives.droid.gui.report.ReportViewFrame.access$200(ReportViewFrame.java:77)
    at uk.gov.nationalarchives.droid.gui.report.ReportViewFrame$2.actionPerformed(ReportViewFrame.java:178)
    at javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:2018)
    at javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2341)
    at javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:402)
    at javax.swing.DefaultButtonModel.setPressed(DefaultButtonModel.java:259)
    at javax.swing.plaf.basic.BasicButtonListener.mouseReleased(BasicButtonListener.java:252)
    at java.awt.Component.processMouseEvent(Component.java:6505)

snail1966:
Hi Gary,
Would you be able to provide any further information including which report this was, and the circumstances e.g. was it for a single or multiple profiles, did you save them first, were there any filters etc. Also, any other relevant information such as the operating system, Java version etc.  
Thanks, Brian

gmcgath:
launch droid.sh
Work with default untitled profile
Add a folder
Click Save and save profile
Click Start and run profile
Define filter: File extension equal to pdf
Click Report, check the saved profile, choose "File count and sizes by mime type", Click "Report on profiles..."
In report, click on "Export"
Navigate to Desktop and name file Problem.txt
Click Save without changing format
Get error described in report

java -version
java version "1.7.0_25"
Java(TM) SE Runtime Environment (build 1.7.0_25-b15)
Java HotSpot(TM) 64-Bit Server VM (build 23.25-b01, mixed mode)

Operating System: OS X Yosemite 10.10.5

[...]

PR #134, opened 2017-02-10
Fix bug #74

rhubner:
Fix bug #74


rhubner:
[#74 Saving report as text in Mac sometimes fails](https://trello.com/c/eIAad7lV/5-74-saving-report-as-text-in-mac-sometimes-fails)

PR #138, opened 2017-02-10
Uri endcoder

### LINKED ISSUE #95: Issue with CSV export percentage coding reserved characters in uri field
The uri field in the CSV export of droid results does not percentage encode some reserved characters. For example + should be encoded as %2B

rhubner:
Fix #95


rhubner:
[#95 Issue with CSV export percentage coding reserved characters in URI field](https://trello.com/c/KAamGoW3/4-95-issue-with-csv-export-percentage-coding-reserved-characters-in-uri-field)

PR #141, opened 2017-03-08
ISO 9660 support

rhubner:
First support for ISO 9660 and Joliet extension. Doroid is able to identify files inside ISO image. ISO images in (g)zip,tar and other archives are not supported at the moment.


rhubner:
[DROID scans contents of ISO](https://trello.com/c/xL0qoxyj/50-droid-scans-contents-of-iso)

sulemanchaudhry:
Change request code reviewed without issue.

Issue #96, opened 2016-07-11
CSV export order currently random


paulyoung84:
The order in which files appear currently in a csv export is random and different each time it is exported. Request that the export is ordered in a fixed way so that each export would be in the same order.

rhubner:
![](https://github.trello.services/images/mini-trello-icon.png) [#96 CSV export order currently random](https://trello.com/c/pD7kDQk5/9-96-csv-export-order-currently-random)

DavidUnderdown:
In test

PR #146, opened 2017-03-27
Sort csv export by URI.

rhubner:
Fix bug #96


rhubner:
[#96 CSV export order currently random](https://trello.com/c/pD7kDQk5/9-96-csv-export-order-currently-random)

PR #153, opened 2017-05-04
Fix PUI properties for ISO container.

rhubner:


sulemanchaudhry:
As per conversation with Radek, happy to accept the merge request.

PR #156, opened 2017-05-24
Add support for new containers in non profile mode.

rhubner:


rhubner:
[Quick fix of new archival container contents ID](https://trello.com/c/tkgOjq52/77-quick-fix-of-new-archival-container-contents-id)

PR #158, opened 2017-05-25
Show root folder as empty when is empty.

rhubner:


rhubner:
[BUG Fix empty folder](https://trello.com/c/VFx01EIs/83-bug-fix-empty-folder)

PR #159, opened 2017-05-31
Fix seven zip(7z) implementation in profile mode.

rhubner:


rhubner:
[7zip broken in profile mode](https://trello.com/c/2PzCZXCK/86-7zip-broken-in-profile-mode)

PR #167, opened 2017-06-26
Dependent library updates and NIO.2 for performance

adamretter:
In the interests of improving performance:

* Updated all Maven plugins to the latest versions
* Fixed problems with dependencies in the pom.xml files
* Updated all dependent libraries to the latest version
* Introduced OWASP into the Maven build to spot security issues with dependent libraries
* Switched DBCP for HikariCP
* Switched from OpenCSV to uniVocity
* Switched all java code (except droid-gui) over to use NIO.2
* All I/O is now buffered to improve performance

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
-dist: precise
+sudo: false
 language: java
 jdk:
- - oraclejdk7
-before_install:
- - "sudo apt-get update"
+ - openjdk7
+ - openjdk8
+ - oraclejdk8
 before_script:
  - "export DISPLAY=:99.0"
  - "sh -e /etc/init.d/xvfb start"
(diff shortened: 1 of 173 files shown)


rhubner:
Wow, great pull req. I will look on it. I'm not sure if we can include this to 6.4.

adamretter:
@rhubner Okay so I fixed the build. I believe that all tests pass now. I also updated the Travis settings, the builds should be faster and are now tested and pass on JDK 1.7 and JDK 1.8.

rhubner:
[RH to write to AR requesting that he split the pull request into smaller chunks and fixes the error in the current version](https://trello.com/c/Ee34SbfR/98-rh-to-write-to-ar-requesting-that-he-split-the-pull-request-into-smaller-chunks-and-fixes-the-error-in-the-current-version)

PR #172, opened 2017-07-19
Filter 'Running DROID.txt' to update droid jar versions.

rhubner:


rhubner:
[Adjust Running Droid.txt for 6.4](https://trello.com/c/dLQZl4rE/84-adjust-running-droidtxt-for-64)

PR #177, opened 2017-08-30
Minor fixes (proposal)

kant:


droid-help/src/main/resources/Web pages/Information collected by DROID.html
```diff
@@ -1,6 +1,6 @@
 <!--
 
-    Copyright (c) 2016, The National Archives <pronom@nationalarchives.gsi.gov.uk>
+    Copyright (c) 2017, The National Archives <pronom@nationalarchives.gsi.gov.uk>
```
adamretter: This needs to be 2016, to be consistent. Otherwise the build will fail at the license check stage.

PR #180, opened 2017-08-30
Minor fixes (proposal)

kant:


alexgreenDP:
Hello kant, 
Thank you for your pull requests. We do not update copyright as it should remain as the date the code was first released - that's why some of the checks are currently failing. If you would like to remove that change from all of your pull requests and merge them into one we will be happy to accept them. Thanks

droid-help/src/main/resources/Web pages/FAQ.html
```diff
@@ -1,6 +1,6 @@
 <!--
 
-    Copyright (c) 2016, The National Archives <pronom@nationalarchives.gsi.gov.uk>
+    Copyright (c) 2017, The National Archives <pronom@nationalarchives.gsi.gov.uk>
```
adamretter: This needs to be 2016, to be consistent. Otherwise the build will fail at the license check stage.

PR #185, opened 2017-09-15
Early termination for match collection

tokee:
In [webarchive-discovery](https://github.com/ukwa/webarchive-discovery) we use droid on billions of files of varying sizes, as part of indexing webpages and other resources. Performance testing on sample files suggests that droid takes up 20-30% of total processing time. We might get by by only using the first signature match returned by droid, which would open up for early termination in the droid signature match logic.

This pull request overloads selected methods with an optional `maxMatches` parameter, which controls this early termination. The code in `ContainerSignatureMatch.matchBinaryContent` has been updated to take advantage of this.

`MatchLimitTest.testLimitSpeed` is a performance test that uses the provided sample files. This extremely rough performance test suggest a doubling of processing speed by using early termination, but the gain (there is never a loss) in speed is highly dependent on the input file.

### DIFF
--- a/droid-container/src/main/java/uk/gov/nationalarchives/droid/container/ContainerSignatureMatch.java
+++ b/droid-container/src/main/java/uk/gov/nationalarchives/droid/container/ContainerSignatureMatch.java
@@ -165,7 +165,7 @@ public class ContainerSignatureMatch {
             Map<String, ContainerFile> sigFiles = signature.getFiles();
             InternalSignatureCollection binSigs = sigFiles.get(entryName).getCompiledBinarySignatures();
             if (binSigs != null) {
-                matched = binSigs.getMatchingSignatures(content, maxBytesToScan).size() > 0;
+                matched = binSigs.getMatchingSignatures(content, maxBytesToScan, 1).size() > 0;
             }
             if (matched) {
                 unmatchedFiles.remove(entryName);
--- a/droid-core/src/main/java/uk/gov/nationalarchives/droid/core/BinarySignatureIdentifier.java
+++ b/droid-core/src/main/java/uk/gov/nationalarchives/droid/core/BinarySignatureIdentifier.java
@@ -92,11 +92,27 @@ public class BinarySignatureIdentifier implements DroidCore {
      */
     @Override
     public IdentificationResultCollection matchBinarySignatures(IdentificationRequest request) {
+        return matchBinarySignatures(request, -1);
+    }
+
+    /**
+     * Submits an identification request to identify files using
+     * binary signatures.
+     *
+     * @param request the identification request.
+     * @param maxMatches The maximum number of matching signatures to collect formats for.
+     *                   Setting this low (i.e. to 1) can speed up processing speed by use of early termination.
+     *                   -1 means unlimited matches.
+     *                   Note that the number of collected formats can exceed maxMatches, as each signature can
+     *                   deliver multiple file formats.
+     * @return the identification result.
+     */
+    public IdentificationResultCollection matchBinarySignatures(IdentificationRequest request, int maxMatches) {
         //BNO: Called once for each identification request
         IdentificationResultCollection results = new IdentificationResultCollection(request);
         results.setRequestMetaData(request.getRequestMetaData());
         ByteReader byteReader = new IdentificationRequestByteReaderAdapter(request);
-        sigFile.runFileIdentification(byteReader);
+        sigFile.runFileIdentification(byteReader, maxMatches);
         final int numHits = byteReader.getNumHits();
         for (int i = 0; i < numHits; i++) {
             FileFormatHit hit = byteReader.getHit(i);
--- a/droid-core/src/main/java/uk/gov/nationalarchives/droid/core/signature/droid6/FFSignatureFile.java
+++ b/droid-core/src/main/java/uk/gov/nationalarchives/droid/core/signature/droid6/FFSignatureFile.java
@@ -494,7 +494,22 @@ public class FFSignatureFile extends SimpleElement {
      * @param targetFile The binary file to be identified
      */
     public final void runFileIdentification(final ByteReader targetFile) {
-        final List<InternalSignature> matchingSigs = intSigs.getMatchingSignatures(targetFile, maxBytesToScan);
+        runFileIdentification(targetFile, -1);
+    }
+    
+    /**
+     * Identify the target file using the signatures defined in this signature file.
+     *
+     * @param targetFile The binary file to be identified
+     * @param maxMatches The maximum number of matching signatures to collect.
+     *                   Setting this low (i.e. to 1) can speed up processing speed by use of early termination.
+     *                   -1 means unlimited matches.
+     *                   Note that the number of collected formats can exceed maxMatches, as each signature can
+     *                   deliver multiple file formats.
+     */
+    public final void runFileIdentification(final ByteReader targetFile, int maxMatches) {
+        final List<InternalSignature> matchingSigs =
+                intSigs.getMatchingSignatures(targetFile, maxBytesToScan, maxMatches);
         final int numSigs = matchingSigs.size(); // reduce garbage: use an indexed loop rather than an iterator.
         for (int i = 0; i < numSigs; i++) {
             final InternalSignature internalSig = matchingSigs.get(i);
--- a/droid-core/src/main/java/uk/gov/nationalarchives/droid/core/signature/droid6/InternalSignatureCollection.java
+++ b/droid-core/src/main/java/uk/gov/nationalarchives/droid/core/signature/droid6/InternalSignatureCollection.java
@@ -93,6 +93,21 @@ public class InternalSignatureCollection extends SimpleElement {
      * @return A list of the internal signatures which matched. 
      */
     public List<InternalSignature> getMatchingSignatures(ByteReader targetFile, long maxBytesToScan) {
+        return getMatchingSignatures(targetFile, maxBytesToScan, -1);
+    }
+
+    /**
+     * Runs all the signatures against the target file,
+     * adding a hit for each of them, if any of them match.
+     *
+     * @param targetFile The file to match the signatures against.
+     * @param maxBytesToScan The maximum bytes to scan.
+     * @param maxMatches The maximum number of matching signatures to return.
+     *                   Setting this low (i.e. to 1) can speed up processing speed by use of early termination.
+     *                   -1 means unlimited matches.
+     * @return A list of the internal signatures which matched.
+     */
+    public List<InternalSignature> getMatchingSignatures(ByteReader targetFile, long maxBytesToScan, int maxMatches) {
         //BNO: intSigs here represents all the available binary signatures..
         List<InternalSignature> matchingSigs = new ArrayList<InternalSignature>();
         if (targetFile.getNumBytes() > 0) {
@@ -101,13 +116,15 @@ public class InternalSignatureCollection extends SimpleElement {
                 final InternalSignature internalSig = intSigs.get(sigIndex);
                 if (internalSig.matches(targetFile, maxBytesToScan)) {
                     matchingSigs.add(internalSig);
+                    if (maxMatches != -1 && matchingSigs.size() >= maxMatches) {
+                        break; // Skip the rest of the signatures as the limit has been reached
+                    }
                 }
             }
         }
         return matchingSigs;
     }
-    
-   
+
     /**
      * Prepares the internal signatures in the collection for use.
      */
--- a/droid-core/src/test/java/uk/gov/nationalarchives/droid/core/MatchLimitTest.java
+++ b/droid-core/src/test/java/uk/gov/nationalarchives/droid/core/MatchLimitTest.java
@@ -1,0 +1,191 @@
+/**
+ * Copyright (c) 2016, The National Archives <pronom@nationalarchives.gsi.gov.uk>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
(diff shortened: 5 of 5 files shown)


droid-container/src/main/java/uk/gov/nationalarchives/droid/container/ContainerSignatureMatch.java
```diff
@@ -165,7 +165,7 @@ public void matchBinaryContent(String entryName, ByteReader content) {
             Map<String, ContainerFile> sigFiles = signature.getFiles();
             InternalSignatureCollection binSigs = sigFiles.get(entryName).getCompiledBinarySignatures();
             if (binSigs != null) {
-                matched = binSigs.getMatchingSignatures(content, maxBytesToScan).size() > 0;
+                matched = binSigs.getMatchingSignatures(content, maxBytesToScan, 1).size() > 0;
```
adamretter: Why `1` here instead of `-1` (which would have been the previous default)?

tokee: It is a speed optimization, using early termination: The overall check is whether there is any match at all (the `> 0` part). As soon as there is 1 hit, that check is satisfied and there is no need to look for further hits.

adamretter:
@Dclipsham What do you think of this one?

nishihatapalmer:
Doing less work will certainly speed things up, but as David says, at the expense of accuracy or sometimes correctness.  It would be hard to predict what the exact effect would be (depends on order of signatures processed, what signatures are prioritised over others).  Limiting matches to a number is a very blunt tool to obtain a speed up.

One thing we did consider a while ago was stopping binary matching once we had a high confidence match.  The question then becomes: what do we consider a highly confident match?

If we know we have a match where all the related priority signatures have been examined already, then we have pretty good confidence in this match (even if it's possible for other completely different signatures to match).

So one confidence heuristic would be: stop processing if we have a binary match where all related signatures have also been processed (all priorities are now resolvable).

This isn't perfect - it's possible to match other signatures too (e.g. a fragment of HTML embedded in another format).  Dealing with that might require some kind of additional signature prioritisation.  Since that would be a subjective judgement, it would require another file format listing these confidence priorities - quite a bit more work.  Not a good idea to hard code that in DROID.

anjackson:
FWIW, one way to handle this is to rearrange the order of the binary signatures so that they are attempted in priority order, rather than finding all matching signatures then working out which one is the highest priority.

DavidUnderdown:
I think that would might need changes to the PRONOM stored procedures that create the signature file, to output the signatures in an optimised way? We'd have to think a bit about what that would actually be: do we somehow cluster similar signatures together and then discount multiple signatures at the same time?

nishihatapalmer:
I think DROID could cluster signatures and order then based on
prioritisation metadata.  I believe it already produces a sort order for
the ByteSequences within signatures, could do something similar for the
signatures themselves.

Would also require the logic to skip a cluster if a high priority one had
already matched.

Would be worth estimating how much processing could be avoided this way.  I
think the majority of signatures don't have priority info.

nishihatapalmer:
Just did some basic stats.

Out of 1535 formats, 298 have some priority over 738 others (a format can
have priority over more than one format).

tokee:
New comments after 16 months. Fetching relevant memory from long-term storage :smiley:

One thing I did consider was a "if this signature matches, these other signatures cannot match"-logic derived from the signatures themselves. Unfortunately determining these rules seems to be extremely difficult for anything but the very simple "file must start with these exact bytes" and ad-hoc testing showed that those rules were also (non-surprisingly) very fast to process, so that the elimination would not result in any substantial speed up.

nishihatapalmer:
In fact, the code already sorts the signatures to try to maximise performance.

InternalSignature has this code:

/* Assign a sort order on the basis of what kinds of byte sequences appear in the signature:
         *  This is currently defined as follows, where:
         *
         *  1      B  = BOF sequence
         *  2      B* = BOF sequence followed by * subsequence
         *  4      V  = Variable sequence potential full file scan from beginning = *B
         *  8      E  = EOF sequence
         *  16     E* = EOF sequence followed by * subsequence
         * 
         * We assign sort order by treating these as bits to group the different sequences together.
         * and to order them in a way that maximises cache hits on the file reads.

nishihatapalmer:
This could obviously be modified to include higher priority signatures first before lower priority ones.

In order to actually improve performance however, it would also have to filter out lower priority signatures dynamically if a higher priority one was matched.

And we'd maybe also want a flag to say: stop further processing once a high priority signature matches.

nishihatapalmer:
The existing code only improves performance by grouping signatures that operate on similar parts of a file together.  DROID does some fairly intelligent caching though, so it's unclear how much performance different this sorting actually makes.

anjackson:
> And we'd maybe also want a flag to say: stop further processing once a high priority signature matches.

Sorry if I've misunderstood, but I think that's what the original request was proposing? The sorting of signatures is an attempt to make this possible without sacrificing accuracy -- if the signatures are sorted from high to low priority, you should be able to safely stop on the first match.

Apologies if I'm misremembering the code dynamics. I've not looked at DROID source for a while.

nishihatapalmer:
Well, there's two possible ways to speed the code up.

If we have a match for a signature that takes priority over another, we can
skip matching the lower priority signatures without any loss of accuracy.
No flag would be needed for that; it would be a general speed up.  Not sure
how much speed up might be possible, but it's there.

The flag would be to say stop processing all further signatures once a
match was found.  Combining these two would ensure that you would never
miss a higher priority signature by stopping early.  But clearly, stopping
matching once any match is found runs the risk of misidentifying a format
(or at least, missing that there was another possible match from an
unrelated (not lower priority) signature.

nishihatapalmer:
In reality, the first option wouldn't result in much speed up.  While there are a lot of high priority signatures, we only get to skip the ones which are directly lower priority signatures than the one we matched.  So at best we'd probably only be skipping less than 10 signatures in the event of a higher priority match.  

It's the "stop matching once we have a match" which could result in a  bigger general speed up.  So I guess your original proposal - sort in order of priority, and stop on any match would probably work just as well.

However - it's not clear there is a full natural ordering of higher priority to lower priority signatures.  It may be a partial order - since any format can be set as having priority over any other.  Have to investigate what the data model really allows here and if there is a natural priority ordering.

nishihatapalmer:
I've been playing with sorting the FileFormats in priority order (topological sort), then sorting the signatures based on the FileFormat priority sort order.  

FileFormats priorities form a directed acyclic graph (not a tree).  They must contain no cycles in the priority relationships (they don't currently, and if they do it's detected by the sort code so no infinite loops).

```
  F1 --->  F2 ---> F3    F6     F7
   |        \      /            |
   |         \   /              |
   v           v                v
  F4 --------> F5               F8
```

This will almost always run Signatures in the order of format priority, meaning if you stopped processing on a first match, you won't accidentally match a lower priority format, as the higher ones will have already run.

However - it can't guarantee it, because signatures can link to more than one file format.  If it links to a high priority format, it must be sorted as a priority.  It it also links to a lower priority format, then a hit on that signature will also flag a hit on the lower priority (potentially before the higher priority signature for it has had a chance to run).

Will keep playing to see if there's a way to at least detect such situations.  I suspect there aren't any.

nishihatapalmer:
I tried matching signatures stopping on a first match, with both format priority sort and without.

Sorting signatures by format priority massively increases the accuracy of matching when stopping on a first match.

But... There are still a substantial number of mismatches Vs matching all signatures.  Not sure what is causing that yet.

nishihatapalmer:
I've identified one source of mismatch.

Where there are both binary signatures and container signatures available, stopping on first match will always use the binary signature and won't go on to process the (possibly more accurate) container signatures.

For example:

fmt/135 has a binary signature for Open Document Format 1.0
fmt/136 has a binary signature for Open Document Text 1.0, and also has a container signature for it.

With format priority sort, and stop on first match, any one of these binary identifications will be processed before the ZIP format it's based on, as they are higher priority.  The zip format will never be identified, and no container matching will be done.

One way to increase accuracy would be to ensure that container signature formats are always processed first (so we know if they're containers, and to keep processing (don't stop on first match for those signatures only).  That way, we'll know they are container signatures, which would then trigger the container signature matching, and the usual format priority calculations would be done.

Of course, this strategy is less performant than just stopping on a first binary match.  One way to mitigate it would be to process container signatures before most of the binary signatures (just try to recognise the two container formats first, and if we have one, move on to full container signature recognition immediately).

nishihatapalmer:
Another way would be to remove binary signatures for which there are perfectly good and generally more accurate container signatures.  

This might affect accuracy slightly (if the container didn't match for some reason but the binary still did), but it would ensure that the more accurate signatures ran in preference to the less accurate ones.

nishihatapalmer:
Having just run some informal profiling, the performance increase as it currently stands when stopping on first match is around 7%.

If I modify it for more accuracy by not missing viable container signatures, that will probably be a bit less.  If I strip out binary signatures for which there is a viable container signature, not sure whether it will be faster or slower (probably slower, as it has to process the container signatures, but only misses a few binary signatures).

nishihatapalmer:
In general when profiling DROID, it is mostly IO bound.  It manages to get around 80% CPU utilisation by running multiple threads, so when when one thread is blocked on IO, another thread can usually make progress.

Nevertheless, IO is still the major limiting factor in how fast DROID can generally go.  If we made signature recognition an order of magnitude faster, I doubt we'd see a big improvement in overall performance.  Maybe we'd get 15% better, or 20%.  The CPU would just be sitting idle for longer waiting for more data to process (so it may use less power, but not get much faster overall).

nishihatapalmer:
So, assuming we tweak things a bit more and manage to get a 10% speed up (at the expense of some accuracy), and we say that DROID occupies 30% of total webarchive discovery processing time, that gives an overall speed up of 3%.  If we get 20% (unlikely) then we get a 6% speed up.

I guess the question is whether the loss of accuracy is worth around 5% speed up in web archive discovery?

nishihatapalmer:
One more thought.  There may be a way to generally speed up matching a bit without any real loss of accuracy.

Currently, DROID

- runs all binary signatures.   
- If one of them was a container trigger format (OLE2 or ZIP) it runs the container signatures for that trigger format.  
- Finally, it removes any lower priority matches. 

An alternative strategy might be:

- Look for the container trigger format signatures first.
  - If we have a container trigger format match, process the container signatures for that format:
    - If we have a container signature match, accept it and don't process binary signatures.
    - If we don't have a container signature match, process all the other binary signatures.
  - If we don't have a container trigger process the binary signatures not dependent on the container trigger formats (i.e. if we didn't find OLE2, no point in processing binary signatures based on OLE2).

nishihatapalmer:
Probably not worth the complexity of removing dependent binary formats - we'd only be looking at a few out of hundreds.

nishihatapalmer:
We could optionally also use the format priority sort I started with when processing the binary signatures *after* we get no container signature match, and stop on a first binary match again.  This would be marginally less accurate, but possibly a bit faster.  It would give better results than just stopping on a first binary match before processing container signatures.

nishihatapalmer:
So, I've implemented the container first strategy in a branch, and centralised all the identification code.  This gives a smallish speed up (<5%) without changing results at all (depending on how many container formats we end up processing).

At some point I'll try using the stop-on-first-match strategy combined with a format priority sort, and see where that gets us.  I suspect results will be almost identical, with the only outliers being ones which identify as multiple formats due to signature problems or a strange-but-legitimate mix of formats.

Performance should improve a little again.  I guess it all adds up - combining both strategies might get us somewhere between 5-10% performance improvement, which would be nice but not earth shattering.  All the little improvements do add up though.

nishihatapalmer:
A question for @tokee ...  How is DROID used in your environment?  Were you invoking via the command line?  If so, were you using the No Profile mode or the Profile mode followed by an export?

These things may influence how much performance improvement we may get.  Another change which may be happening is to unify the command line with the rest of DROID.  The No Profile mode is currently single-threaded - but when it is unified, it will use multi-threaded profiling.

Dclipsham:
Noting @nishihatapalmer has offered alternative approaches to optimisation, I'm proposing to close this un-merged. Please shout if you feel strongly otherwise...

tokee:
@nishihatapalmer I just got a ping from @Dclipsham s comment. Sorry for not noticing your question earlier. The relevant Droid.using code is [DroidDetectorAnalyser](https://github.com/ukwa/webarchive-discovery/blob/master/warc-indexer/src/main/java/uk/bl/wa/analyser/payload/DroidDetectorAnalyser.java) from the `webarchive_discovery`-project. It is very simple: Just create a `DroidDetector` (possibly using binarySignaturesOnly, but that is false per default) and call `DroidDetector.detect`, which returns the first match from the list of matches.

As for closing this pull-request, I am too far removed (in time as well as focus) to have any strong feelings for this. I thought the pull request was "neutral", by which I mean that it did not change behaviour unless asked to, but it seems that I was mistaken. As such I fully understand not wanting to accept it.

nishihatapalmer:
I think this can be safely closed at this point.  I've done a lot of experimenting with early match termination, and any matching performance advantage in DROID is generally swamped by IO.  

There are potential small benefits, but they can only be realised by adding a priority sort to signatures to maintain matching accuracy, along with some other useful optimisations like matching container signatures first.  

We already have a "match container first" PR (which will need some updating anyway due all the changes in the latest release.  It would be reasonably easy to build on that once updated to add this functionality.  

So this PR doesn't really have what is needed to make that happen - it causes a huge loss of matching accuracy, and we should close it, but I already have much of what would be needed to add the basic idea later without major loss of accuracy.

PR #187, opened 2017-10-04
Move droid to Java 8

adamretter:
This PR moves DROID to require Java 8 as a minimum. It also updates DROID to Spring 5; which is a pre-requisite to enabling support for Java 9 platforms.


rhubner:
[Move droid to java 8](https://trello.com/c/fy7YDOsr/110-move-droid-to-java-8)

PR #190, opened 2017-10-19
Exclude appveyor.yml from licence header check.

rhubner:


rhubner:
[Build droid on appveyor.com](https://trello.com/c/0JDyPry2/97-build-droid-on-appveyorcom)

PR #207, opened 2018-05-16
Changed apache logging/java utils logging to slf4j api

Urborging:
Squashed commits to comply with comments on previous pull request

### DIFF
--- a/droid-command-line/pom.xml
+++ b/droid-command-line/pom.xml
@@ -162,6 +162,10 @@
             <scope>runtime</scope>
   	    </dependency>
         <dependency>
+            <groupId>org.slf4j</groupId>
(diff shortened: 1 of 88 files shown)


rhubner:
Hello Adam,

In droid-ui we havae this dependency : 

'''xml
<dependency>
  <groupId>org.slf4j</groupId>
  <artifactId>slf4j-log4j12</artifactId>
</dependency>
'''

The goal is, to have only slf4j API dependency in all modules except the droid-ui. I should allow easy integration to other software.

adamretter:
@rhubner yup I saw that. But won't you also need the same for `droid-command-line` as well as a log4j config file for each (which will also need to be packaged by `droid-dist`?

PR #211, opened 2018-07-12
Cache maven artifacts in travis builds.

rhubner:


adamretter:
Interesting... does this make any noticeable difference in build times?

rhubner:
Hello @adamretter 
Yes, it does. Sometimes maven is stuck to download dependency from maven central. Also it reduce load on maven central repository.

[This build](https://travis-ci.org/digital-preservation/droid/builds/403134912) is without cache, you can see that one build took 19 min. What I saw during buil, most of the time was slow download from maven central. Also travis-ci suggest this in [official documentation](https://docs.travis-ci.com/user/caching/).

Radek

PR #222, opened 2018-11-04
Let archive files be embedded in any other archive file.

nishihatapalmer:
The existing code restricts many archive types such that they can only be read if they are directly inside a file.  This pull request uses interfaces supported by those archive types, and adapts WindowReaders to them.  This allows us to process these archive types no matter where they appear - in files, embedded in zips, tar.gz, or any nested combination.

Note that this is not a new approach - it's exactly what already exists to support TrueZip reading.

 * New classes to adapt WindowReaders to the various interfaces supported by different archive types.
 * Tests for the new classes (and a test for the existing TrueZip reading classes).


nishihatapalmer:
Oops - need to make some further changes - closing for now.

nishihatapalmer:
Hmmm... it builds fine on my machine, but AppVeyor has problems with it.  Re-opeining to see what the issues are, then I'll close again.

nishihatapalmer:
Forgot to commit a change to another package.  Will close and resubmit.

PR #223, opened 2018-11-04
Let archive files be embedded in any other archive file.

nishihatapalmer:
The existing code restricts many archive types such that they can only be read if they are directly inside a file. This pull request uses interfaces supported by those archive types, and adapts WindowReaders to them. This allows us to process these archive types no matter where they appear - in files, embedded in zips, tar.gz, or any nested combination.

Note that this is not a new approach - it's exactly what already exists to support TrueZip reading.

- New classes to adapt WindowReaders to the various interfaces supported by different archive types.
- Tests for the new classes (and a test for the existing TrueZip reading classes).


nishihatapalmer:
The one remaining archive type that is still limited to files as a source is ISO.  

The last release of java iso tools only supports reading from files.  The latest SNAPSHOT code in master does have an interface we could use, but this hasn't been released.

I'm not sure this code is being maintained anymore, very little activity for the last 2 years, and no response to questions about a new release.

One option would be to fork java iso tools and release it, so we could use the SeekableInput interface.

LauraDamianTNA:
Yes, oraclejdk10 is no longer supported. Thanks for the pull request, we'll merge soon

PR #235, opened 2019-08-27
Fix: test not convenient when building in non english environment

JSLair:
- replace "All Files" by the description of the default AcceptAllFileFilter (for example in French it's "Tous les fichiers")


adamretter:
Thanks for your contribution @JSLair

PR #283, opened 2019-10-08
shifting support forward to Java 11 and fix/update dependencies

jcharlet:
- fixed dependency org-openide-util that prevented to use openJdk11 and used official repository rather than outdated TNA third party repository
- set default java version from 1.8 to 11 (most of the work had been done by Radek)
- did a series of non-breaking updates (or requiring small changes) on dependencies, using github dependabot and maven dependency tool

Update 10/10:
won't change default version to Java 11, still keep compatible with jdk 8 to 11 and updated dependencies

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ matrix:
     - jdk: openjdk10
     - jdk: openjdk11
     - os: osx
-      osx_image: xcode10
+      osx_image: xcode10.1
       env: JAVA_HOME=$(/usr/libexec/java_home)
 
 services:
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ matrix:
     - os: osx
       osx_image: xcode10.1
       env: JAVA_HOME=$(/usr/libexec/java_home)
+    - dist: bionic
 
 services:
   - xvfb
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,10 @@
 version: '{build}'
+install:
+  - cmd: SET JDK_HOME=C:\Program Files\Java\jdk11
+  - cmd: SET JAVA_HOME=C:\Program Files\Java\jdk11
+  - cmd: SET PATH=%JDK_HOME%\bin;%PATH%
+  - cmd: mvn --version
+  - cmd: java -version
 build_script:
 - cmd: mvn compile
 test_script:
--- a/droid-core-interfaces/pom.xml
+++ b/droid-core-interfaces/pom.xml
@@ -212,7 +212,7 @@
         </dependency>
         <dependency>
             <groupId>org.hamcrest</groupId>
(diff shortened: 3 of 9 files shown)


adamretter:
@jcharlet You might also like to update `.travis.yml` and `appveyor.yml` ;-)

jcharlet:
> @jcharlet You might also like to update `.travis.yml` and `appveyor.yml` ;-)
> 
> I think if you switch to `bionic` on Travis, then OpenJDK 11 is the default.

yes I noticed that @adamretter :) , it's using java 10 at the moment. and brilliant, thanks for the advice, will check when I'm back tomorrow!

.travis.yml
```diff
@@ -7,7 +7,7 @@ matrix:
     - jdk: openjdk10
```
adamretter: You will need to drop the lines:
```
- jdk: openjdk8
- jdk: openjdk9
- jdk: openjdk10
```

You could also add `dist: bionic`.

jcharlet: hi Adam, 
thanks again for your comments.
just discussed it with David and Laura, David wants us to keep supporting jdk8 for other products embedding droid. So I'll have to keep those.
just tried making new changes, will clean history if it goes through and replace
```
    - os: osx
      osx_image: xcode10.1
      env: JAVA_HOME=$(/usr/libexec/java_home)
```
with 
```
    - os: linux
      dist: bionic
```

jcharlet: ah you probably want to keep the osx build, ok got it, will just add it on top

adamretter: So I don't understand what's happening then. You want to keep JDK 8 support, but you are moving to Java language level 11. My experience is that even if you back-compile, these things are not 100% compatible. Have I missed something?

jcharlet: @adamretter shall we keep osx in the list ? or we can remove it and only use bionic?

Dclipsham: in previous DROID releases we've attempted to maintain backwards compatibility for Java where possible, as some users and particular instances that bundle DROID (preservation systems, toolkits etc) may have limitations on what they can install and may be tied back to earlier Java versions. /if possible/ I'd like to ensure 6.5 can run on 1.8, but need to understand whether this is possible or not.

jcharlet: yeah right :)). Well I think there was some confusion regarding the task "shifting support forward to Java 11", which was already done by Radek. I guess the whole idea from David was more to update all dependencies.
I moved back to Java language level 8.
I'll rename the PR too

jcharlet: @Dclipsham that will work fine on java 8!

jcharlet: as discussed, we won't remove support to java 8. Will resolve this conversation

jcharlet:
builds are passing on CI, and PR is now ready for review :)!

Issue #231, opened 2019-07-29
Null Pointer exception when exporting profile with extension mismatch filters turned on


Dclipsham:
Steps to reproduce (6.4 - both *nix and Windows environments):
1. run droid scan as normal
2. turn filter on by Filter menu> choose extension mismatch filter, hit apply
3.  Export > choose the profile, and select destination

Error appears

Dclipsham:
![FilterNullPE](https://user-images.githubusercontent.com/2189778/62055380-6c250900-b213-11e9-829a-f402dc1717c8.png)

PR #287, opened 2019-10-15
fix: NPE on export when extension mismatch filters turned on 

jcharlet:
fix Null Pointer exception when exporting profile with extension mismatch filters turned on 
https://github.com/digital-preservation/droid/issues/231


jcharlet:
Thanks @adamretter !

adamretter:
No, thanks you @jcharlet

PR #289, opened 2019-10-17
bundle jre with windows

jcharlet:
Provide now 2 packages:
- one for windows with jre 1.8 embedded
- one for unix (osx / linux) without (like it used to be)

Reasoning:
- use Java 8: there is still a small graphical bug when compiled with Java 11 mentioned in another issue that needs fixing, so I'll stick to Java 8 right now.
- started preparing bundles for osx/linux/windows. but It requires to download 3 JREs, create 3 packages, all significantly larger. We agreed with David that users would probably not have any issue having/installing Java if they are on linux/osx, while users on Windows are more likely to have restricted rights.

What do you think @adamretter ? Are you happy with that solution? I can otherwise do a few more commits to include a linux and an osx packages. Will need some help testing the mac version too :).
I did win64 btw, do we need win32? A bit concerned about creating loads of packages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,8 +36,8 @@ You will need JAVA 8 to 11 installed to run it.
 
 once unpacked, use the `droid.sh` script to run the application
 #### Windows users
-Archive  `droid-binary-${VERSION}-bin-win64-with-jre.zip`
+Archive  `droid-binary-${VERSION}-bin-win32-with-jre.zip`
 
-For Windows users who might not be able to install JAVA, the provided bundle includes JAVA 8.
+For Windows users who might not be able to install JAVA, the provided bundle includes JAVA 11.
 
 once unpacked, use the `droid.bat` script to run the application  
--- a/README.md
+++ b/README.md
@@ -8,6 +8,8 @@ The latest binary file can be downloaded from [The National Archives website](ht
 
 More information can be found on the DROID github pages here: http://digital-preservation.github.com/droid/
 
+## General Information
+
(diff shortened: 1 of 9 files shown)


adamretter:
@jcharlet Sounds reasonable to me.

The thing will be making sure that when you have produced runs with the Bundled JRE and not a JRE which is already installed on the system, getting that right can be a bit tricky sometimes ;-)

jcharlet:
> @jcharlet Sounds reasonable to me.
> 
> The thing will be making sure that when you have produced runs with the Bundled JRE and not a JRE which is already installed on the system, getting that right can be a bit tricky sometimes ;-)

I download the jre from adoptopenjdk with ant maven plugin, then include it with the assembly plugin, no risk using the local jre!

Ok then, solution agreed, I'll work on fixing the travis build.

jcharlet:
I don't understand why travis job can't download the jdk, and the logs do not say anything helpful, I want to ssh into it and try running things manually to see why it's stuck.

@adamretter do we have debug mode available on travis job, is it something you could grant me access? https://docs.travis-ci.com/user/running-build-in-debug-mode/

Otherwise can I send a request to travis support team to enable it for our repository?

adamretter:
I would suggest not using the Maven Ant plugin for the download, but rather to use [maven-download-plugin](https://github.com/maven-download-plugin/maven-download-plugin) - it should give you clearer error messages.

jcharlet:
that worked, thanks @adamretter !

Issue #219, opened 2018-09-28
Last modified dates inside a 7zip file


jennymitcham:
When DROID looks in a 7zip file it reports all of the last modified dates as being today's date, rather than accessing the last modified dates for each file.

Dclipsham:
bug, easy to replicate - to be investigated target resolution 6.5

jcharlet:
Fixed

PR #303, opened 2019-11-05
added lastmodified for files inside ziparchive

LauraDamianTNA:
In a similar way it's done in other containers


jcharlet:
Fix https://github.com/digital-preservation/droid/issues/219

jcharlet:
validated on sample .7z file

Issue #237, opened 2019-08-30
full implementation of standard signature syntax in container signatures


Dclipsham:
Container signatures don't currently work when providing certain valid binary signature patterns, e.g. either/or patterns (xx|xx) and inline wildcard ranges {0-4}. Current workarounds involve multiple signature blocks, which is unsatisfactory

nishihatapalmer:
This is a consequence of the original split between PRONOM and DROID.   PRONOM itself compiles the binary signature syntax into a set of instructions to match sequences, with left and right fragments used for alternatives and wildcards.

What DROID really needs is a PRONOM syntax signature parser / compiler.  It won't be needed for binary signatures (since PRONOM is already doing this), but it is needed to enable the full DROID syntax for containers, since PRONOM doesn't get involved in compiling those.

I think this would be a really good idea.  It then frees DROID a hard PRONOM dependency.  If also applied to binary signatures, it would allow anyone to develop DROID signatures using the natural syntax, rather than having to become deep technical experts in how PRONOM works.

nishihatapalmer:
I may have a go at writing this - I've done a lot of work on parsing and compiling in byteseek.  Unless someone else is already working on this?

nishihatapalmer:
A longer term consequence of this would be PRONOM would no longer need to be able to process the signature syntax, just store the signature syntax.

This would make maintaining PRONOM cheaper, and allow the syntax to evolve independently of the registry used to manage the signatures.

It's clearly right that the tool which processes the signatures should decide how they need to be processed and compiled.  

In some cases, PRONOM already produces sub-optimal compilation.  There is already code in DROID to try to mitigate PRONOM's sub-optimal decisions!

ross-spencer:
[...]

PR #304, opened 2019-11-05
Droidsyntaxparser

nishihatapalmer:
These classes can parse PRONOM and container signature syntax, producing an Abstract Syntax Tree (AST).   A compiler for the AST produces ByteSequence, SubSequence and SideFragment objects that can be used in DROID to search for those signatures.


jcharlet:
solves https://github.com/digital-preservation/droid/issues/237

jcharlet:
I close this PR, created a new one directly from this the core repository's branch so that we can both work on it @nishihatapalmer .
all following discussion to happen there: https://github.com/digital-preservation/droid/pull/305

Issue #308, opened 2019-11-10
BUG: setting IdentificationMethod filter results in exception.


nishihatapalmer:
If you set an Identification Method filter, you get the following exceptions:

> 11:17:31,228 ERROR [AWT-EventQueue-0] JDBCProfileDao:334 - A database exception occurred finding filtered nodes with parent id NULL
> org.springframework.dao.DataIntegrityViolationException: PreparedStatementCallback; SQL [select profile.*, case when (profile.identification_method IN (?)) then 1 else case when profile.resource_type <> 2 and exists ( select children.node_id from profile_resource_node as children where children.prefix > profile.prefix and children.prefix < profile.prefix_plus_one and (children.identification_method IN (?))) then 2 else 0 end end as FilterStatus from profile_resource_node as profile where profile.parent_id is null]; An attempt was made to get a data value of type 'INTEGER' from a data value of type 'uk.gov.nationalarchives.droid.core.interfaces.IdentificationMethod'.; nested exception is java.sql.SQLDataException: An attempt was made to get a data value of type 'INTEGER' from a data value of type 'uk.gov.nationalarchives.droid.core.interfaces.IdentificationMethod'.
> 	at org.springframework.jdbc.support.SQLErrorCodeSQLExceptionTranslator.doTranslate(SQLErrorCodeSQLExceptionTranslator.java:247)
> 	at org.springframework.jdbc.support.AbstractFallbackSQLExceptionTranslator.translate(AbstractFallbackSQLExceptionTranslator.java:72)
> 	at org.springframework.jdbc.core.JdbcTemplate.translateException(JdbcTemplate.java:1443)
> 	at org.springframework.jdbc.core.JdbcTemplate.execute(JdbcTemplate.java:633)
> 	at org.springframework.jdbc.core.JdbcTemplate.query(JdbcTemplate.java:669)
> 	at org.springframework.jdbc.core.JdbcTemplate.query(JdbcTemplate.java:700)
> 	at org.springframework.jdbc.core.JdbcTemplate.query(JdbcTemplate.java:712)
> 	at org.springframework.jdbc.core.JdbcTemplate.query(JdbcTemplate.java:768)
> 	at uk.gov.nationalarchives.droid.profile.JDBCProfileDao.findProfileResourceNodes(JDBCProfileDao.java:319)
[...]

PR #310, opened 2019-11-11
FIX: exception on setting enum based filter.

nishihatapalmer:
* Instead of returning the enum object itself, we must return it's
    ordinal value, as the database has an integer field.

### DIFF
--- a/droid-results/src/main/java/uk/gov/nationalarchives/droid/profile/FilterCriterionImpl.java
+++ b/droid-results/src/main/java/uk/gov/nationalarchives/droid/profile/FilterCriterionImpl.java
@@ -219,13 +219,13 @@ public class FilterCriterionImpl implements FilterCriterion {
                 o = ISODateTimeFormat.date().parseDateTime(s).toDate();
                 break;
             case IDENTIFICATION_METHOD:
-                o = IdentificationMethod.values()[Integer.valueOf(s)];
+                o = IdentificationMethod.values()[Integer.valueOf(s)].ordinal();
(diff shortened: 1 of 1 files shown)


nishihatapalmer:
This fixes https://github.com/digital-preservation/droid/issues/308

nishihatapalmer:
You are right, this is fragile.  It's also an antipattern which is repeated across quite a lot of DROID.

I'll have a look at giving the enums a fixed ID, but a quick look shows a lot of places in DROID this happens.

nishihatapalmer:
OK - this isn't a trivial change.

It's easy enough to give the enums in that method a "getID()" method and assign a permanent one on construction.  

But then it has to match what the SQL puts in the database, and this uses methods generic to enum (i.e. ordinal()) to convert an enum value into an int to store in the db.  

Conversion the other way around needs similar treatment, and all the filter dialogues and other things that touch this would need to be updated so they all use the same values.

The biggest change would be to the SQL conversion code, as all it knows is it's got an enum, not a whole load of different enums all with getID() methods on them...

adamretter:
@nishihatapalmer okay, so I will merge it as is.

nishihatapalmer:
@adamretter It's a good point though - this is really fragile.  I'll raise an issue so we don't forget about it.  It's something that should probably be done, but quite carefully!

nishihatapalmer:
Raised issue: https://github.com/digital-preservation/droid/issues/325 to cover the use of ordinal values in DROID in general.

PR #312, opened 2019-11-11
FIX: NullPointerException on Report / Export... button click.

nishihatapalmer:
* Path can be null, but was just doing lastSelectedDir.toFile().
  * Just add null test and pass null if so , rather than trying to get tofile().


nishihatapalmer:
Fixes issue: https://github.com/digital-preservation/droid/issues/309

nishihatapalmer:
I really feel I have to comment that many of the checkstyle rules used by DROID are unnecessarily fiddly.  

I've never worked on a project with such restrictive style rules - and yet a lot of code in DROID is still quite horrible (despite conforming to the strict style rules).

They cause considerable friction writing code for DROID, but in many cases don't add in any meaningful way to the readability or maintainability of it.

adamretter:
@nishihatapalmer I feel your pain. Checkstyle was there way before I started contributing. Instead of checkstyle it would be better to add an automated formatter IMHO.

PR #317, opened 2019-11-19
upd: binary bundle without jre now works for windows too

jcharlet:
Enable windows users to also use their own java JRE to run droid.


jcharlet:
agreed with @Dclipsham

Issue #306, opened 2019-11-08
BUG: folders are messed up in UI on latest master.


nishihatapalmer:
If you do a scan in the DROID UI after building the latest master, it all seems to work.

But if you open folders in the UI, the tree goes crazy, with different folders opening, elements at different levels inside the same folder.

It seems to work OK with one sort order of the Resource column, but mostly it's messed up.  Not sure if there have been any changes to the UI code recently.

nishihatapalmer:
I did a git clone of the latest master and built it from scratch.  Wasn't sure if this was something to do with any changes I've been fiddling with myself - but it is definitely a problem in master.

nishihatapalmer:
I wonder if it has anything to do with updating the org-netbeans-swing-outline dependency in the commit : https://github.com/digital-preservation/droid/commit/431e4b490940f55b8dbe6a0abcb5654910f429d7#diff-e18fa3c09c9a75e713b7455cc2b7e3e4

I think that's what draws the tree in DROID.

nishihatapalmer:
This commit is the cause of the tree drawing issues.

I reverted the change and put the libs back to the 7.2 versions.   With just that change, all is back to normal.

nishihatapalmer:
(I reverted all the netbeans libraries, not just that one, back to 7.2)

nishihatapalmer:
Note: while reverting the change fixes the issue, there may be a benefit to updating to a later release.  It may be easy to fix, but we should revert this change until we understand how to make it work in the updated version.

jcharlet:
@nishihatapalmer can you please tell me on which jdk (open / oracle) and which version (8/11/?) are you compiling droid?

I get the following error when compiling on openjdk 11


```
[...]

PR #319, opened 2019-11-19
Fix folders messed up in UI

jcharlet:
PR to fix https://github.com/digital-preservation/droid/issues/306 based on PR https://github.com/apache/netbeans/pull/1639
related to netbeans bug: https://issues.apache.org/jira/browse/NETBEANS-3401


nishihatapalmer:
This fix works, but I think the alternate fix in https://github.com/digital-preservation/droid/issues/306 is probably better, by altering the TreePathSupport class.

It ensures that the TableModelEvent fired by the Outline knows about the new children (makes the EventBroadcaster the last listener, so all tree updates are complete by the time it processes it).

jcharlet:
my bad, I'm so switching between 3 PR, that I forgot about that, sure, will do that change instead, thanks for reminding @nishihatapalmer

Issue #229, opened 2019-06-19
Nested folders in OLE2 container


thorsted:
It appears the code for OLE2 only looks at the root of the document and can't parse files in nested directories unlike the ZIP engine.

DavidUnderdown:
Do you think this would be useful behaviour, I think we're trying to make sure we identify the "top-level" format (since eg a Word doc could have Excel tables embedded etc).  I don't think we've intended to try and identify every component, whereas in a zip file it's typically the files inside that are really of interest.

thorsted:
I currently have two formats I am working on now which would require this behavior. One is an OmniPage OPD file which has no files at the root only folders, another is a Microsoft Home Publishing file which is mostly identical to a MSWorks 5 file, but with a unique folder within the structure. In both cases I would need to identify a file or folder within the internal directories like a ZIP.
[Samples.zip](https://github.com/digital-preservation/droid/files/3320730/Samples.zip)

Dclipsham:
This is definitely required behaviour. The SIARD 2.1 format identification wouldn't work without this functionality existing for ZIP-based containers, and this issue would bring OLE2 containers in line and facilitate more complex container signature patterns for OLE2 types. It wouldn't adversely affect existing signatures that are using files at root level.

nishihatapalmer:
There's a PR for this ( https://github.com/digital-preservation/droid/pull/321 ) which adds the ability to scan all sub-folders and files of an OLE2 container.

I do not have a good container signature using paths to test it with.  It certainly iterates over all the sub-folders and files inside an OLE2, but I can't swear it all works until someone tries a real signature with it.

Paths are specified just like in ZIP container files, using / to separate folders.  The root does not have a starting /.

For example:

```
File1
File2
Directory1
Directory1/File3
Directory1/File4
Directory2
Directory2/File5
```

thorsted:
Excellent work, This is the format I was trying to write a container signature for when I came across the issue. OmniPage 10 format only has folders and versions 12 & 18 have a Data file inside a Version folder. https://github.com/thorsted/pronom-research-week-2019/tree/master/OmniPage

nishihatapalmer:
I'm afraid I don't have the tools to explore inside OLE2 objects handy to try to build a test signature.  Can you recommend a tool which can dump the internal streams out to files?  I'm on linux, can use Windows at a pinch but it's not set up for dev work.

Dclipsham:
Hi Matt,
[...]

PR #321, opened 2019-11-20
ENHANCE: match signatures in OLE2 sub folders

nishihatapalmer:
* https://github.com/digital-preservation/droid/issues/229
   * Walks over all internal files inside an OLE2, not just the
     immediate children of the root.
   * Paths are separated by / just like ZIP.


adamretter:
@nishihatapalmer thanks this looks great! Would you be able to add some unit tests too please?

nishihatapalmer:
I honestly don't think I have time to write unit tests for this - I'm starting a new job on Monday, so my free time will disappear.

Well - the unit tests would actually be very simple - they could only test that a container signature with sub files or folders actually matches a file that has those properties.  So the code is easy.

The hard part is developing (preferably more than one) container signatures that matches actual files (and doesn't match others), so that the engine can run over them to validate they work.  I have no experience developing signatures, and nor do I have suitable files or tools to do this with.

If there's some clever way to mock objects to avoid having to make real signatures for real files, then I guess that might work (although we'll still need test files with the right properties).  I'm not too up on how mocking works (and I'm not sure it solves this problem).

nishihatapalmer:
I can see on the issue there's a sample container file signature for an actual OLE2 file.  I'll have a look at using this and see how I go - but it should test more than one signature preferably.

nishihatapalmer:
I can confirm that a manual test of this using the Omnipage container signatures and documents linked to in the associated issue works.  All documents are identified as both Omnipage 12 and Omnipage 18.

nishihatapalmer:
Well, the ending path separator issue is certainly there, but that's not what's causing this to fail.

Looks like the signature files provided don't recognise the OLE2 base format, so we never trigger the OLD2IdentifierEngine in the first place.  I'll see if I can hack together a working set of signature files from them all.

nishihatapalmer:
My bad, I'd downloaded the HTML page instead of the actual files...

nishihatapalmer:
Well, there are bugs in this code still.   The actual walking of the internal objects works, but it fails to get a document stream, as the call we're using on the POIFSfilesystem only gets streams from children of the root.

Have to modify our code so we can get streams from any part of the POIFSfilesystem.

nishihatapalmer:
Latest code fixes the streaming issue, and we get a container signature identification for the Omnipage files using internal paths inside OLE2.   

Still need to resolve the path ending issue (I modified the signature files to strip out ending /) - but this may not be the right solution.

nishihatapalmer:
I'll try to find the time to do a bit more manual testing, and I'll have a think about the path ending / issue.

Not sure I'll find the time to automate all of this, but at least we'll have manual confirmation that the code works.

nishihatapalmer:
Latest code matches container signatures including internal paths.

Also matches regardless of whether a directory has a trailing / or not (people do both in practice).

nishihatapalmer:
Added unit tests for OLE2.

Turns out all the existing unit tests were flagged as @Ignore

Everything was throwing null pointer exceptions.  This turns out to be a result of using uninitialised objects.  It's fine normally in DROID as spring sets up all the internal dependencies, but if you constructed objects programmatically, lots broke.

I've made the small changes required to ensure that internal objects are not null when constructed programmatically.   The existing tests (bar one) have been re-enabled, and I've added a simple test that looks for sub-folders and files inside container signatures.

It would be good to add a test for a container signature with internal binary sigs, but the current code makes that quite hard.  Once we get the new signature compiling capabilities (in another PR), this could be added quite easily later.

jcharlet:
so it depends on https://github.com/digital-preservation/droid/pull/305 which is almost ready

jcharlet:
ci failed cause of bug on Travis. Fixed on master, will merge onto this branch to fix it.

jcharlet:
tested locally on the sample file from https://github.com/digital-preservation/droid/issues/229#issuecomment-504984316 
but it doesn't seem to work. it should be expanding the opd file right @nishihatapalmer ?

image

image

nishihatapalmer:
@jcharlet No - it shouldn't expand - it's not an archive handler, it's a container file format (the internal "files" are used to identify the overall parent format, not to expand the entries within them.

The signature I used identified the version 12 and 18 Omnipage formats.  I didn't try it on an Omnipage 10 file.  Does it work on 12 or 18 opds?

nishihatapalmer:
Did you put in the new signature files too?   The default signature files have no support for Omnipage.  There's a binary and container signature file stored along the sample opd files linked to from the issue.

jcharlet:
yes included the new signature

so what's the expected output?

image

image

nishihatapalmer:
The expected output is that the files are identified as Omnipage 12 or 18 files.  This doesn't seem to be working for you.

One of the unit tests in Ole2RootFileTest (testIdentifyOLE2ContainersWithInternalPaths) tries to identify using internal paths (but crucially, doesn't run any binary signatures inside them).  Does that unit test work for you?

nishihatapalmer:
You aren't using the Omnipage container file, only the binary file.

Use the Omnipage container file instead and it should work.

nishihatapalmer:
I mean - use the Omnipage binary file and the Omnipage container file.  Both are needed to identify all the samples.

jcharlet:
image

a bit better, does work on 12 and 18! 
Although omnipage10 is detected as version 12, and omnipage15 as 18. Trying to investigate why

nishihatapalmer:
The container signatures only define version 12 and version 18.  I guess
version 10 matches the version 12 signature and version 15 matches the 18
one.

So the signatures could be improved - but I believe they are working as
specified.

jcharlet:
Yes, ok makes complete sense. So that PR is complete. I just need to fix the CI, and will be able to merge after

thorsted:
Thank you guys for the work on this. The signature appears to work as intended, from the image sent to me, There may be more work needed to define versions better, but I was unable to test with current functionality. I look forward to updating the signature before it is published in PRONOM.

jcharlet:
@thorsted if this can help for your test, I can send you DROID's packaged binary. After as you probably know, you just need to put the binary and container files in the right folder, select it in droid's settings (can show you how if needed).

thorsted:
@jcharlet I would love to test out the updated binary.

jcharlet:
@thorsted https://we.tl/t-6fDDuxQ9FL here you go

thorsted:
@jcharlet Tested and works as expected. Looking forward to adjusting other signatures to take advantage of this enhancement. Let me know when there is a snapshot ready for https://github.com/digital-preservation/droid/pull/322 as well. Need to submit OLM signature. Thanks!

Issue #232, opened 2019-08-21
ZIP Container error


thorsted:
Attempting to make a signature for a file format in a ZIP container. Created signature and tested in DROID and file is only identified as ZIP signature. Checked the log and can see an error: "WARN Could not process the potential container format (ZIP): file:/Users/thorsted/Documents/file.olm	ZIP file spanning/splitting is not supported!"

Checked multiple samples for file format and getting the same error. Is this a bug?

anjackson:
I found this potentially related issue: https://github.com/digital-preservation/droid/issues/71

Are you running the latest version of DROID?

Dclipsham:
Hi Tyler,

Are you able to share an example file that encounters this issue, either here or privately?

David

thorsted:
Running version 6.4. It does seem connected to issue #71. Samples sent to David.

Dclipsham:
Thanks Tyler, easy to reproduce here, and confirming as bug. 
My instinct is that it is similar to #71 and probably has a similar resolution (updating/changing zip handler libraries) - I've revisited the problematic files from #100 and they work with DROID 6.4 but should also be tested as part of any fix for this issue.

Dclipsham:
To note also, These zips unpack happily in 7zip and Windows Explorer, however when browsing the contents before unpacking, both tools have issues gathering properties of the contents, as per the images below
![incompleteProperties7Zip](https://user-images.githubusercontent.com/2189778/63583041-1d5c5c00-c592-11e9-88e9-c3b3d7d527eb.png)
![incompleteProperties](https://user-images.githubusercontent.com/2189778/63583043-1d5c5c00-c592-11e9-85e4-d2c46171df83.png)

thorsted:
David, I also was able to confirm, by recompressing with zip and naming as
OLM, my container signature identifies them correctly.

-Tyler

Dclipsham:
There's been a conversation around this issue on Google Groups - https://groups.google.com/forum/#!topic/droid-list/N1tE3ZuEDbo - it seems that the zip handler either needs updating or replacing as the error handling throwing the warning appears to be unhappy with what should be valid Zip64 structure (https://groups.google.com/d/msg/droid-list/N1tE3ZuEDbo/wWDP_FjYAwAJ)

jcharlet:
hi @thorsted 
my solution is actually not satisfying, I forced droid to identify olm files as archives and scan their contents, while actually it should just be identified as a container file. 

[...]

PR #322, opened 2019-11-21
Fix zip olm

jcharlet:
fix https://github.com/digital-preservation/droid/issues/232 
This hotfix and workaround works on 5 separate olm containers.

Note: the GUI bug previously investigated with @LauraDamianTNA was unrelated and was fixed on https://github.com/digital-preservation/droid/pull/319 .


jcharlet:
image
looking at the output, I actually wonder if that fix is satisfying. Because it describes it as a ZIP format, which it isn't really, it looks like a zip64 but isn't. 
@Dclipsham @thorsted  what do you think?

jcharlet:
here is the binary with the fix: https://we.tl/t-js55uauVCP

jcharlet:
cancelling this PR after discussion with David. We want to treat olm files as containers (and single items) rather as archives and try processing their contents. In that case it's not needed, the thing we need here is a signature.

Issue #204, opened 2018-05-08
Ability to drag and drop files into DROID (inc Mac)


ghost:
This feedback was given to me by one of our Mac users- kp.

nishihatapalmer:
I've submitted a PR that enables file drag drop - not yet tested on Mac or Windows.

jcharlet:
merged PR after testing it on linux. will test it on windows too asap.
can someone test it on mac if I provide binaries?

LauraDamianTNA:
I'll test it on Mac now.

LauraDamianTNA:
It works fine on Mac OS X Yosemite.

PR #324, opened 2019-11-21
ENHANCE: adds drag drop support for files into a profile GUI window,

nishihatapalmer:
Allows you to drag files and folders into a new profile rather than having to use the selection dialog.

Tested on Ubuntu 16.04 - needs testing on Windows and Mac.

Also see https://github.com/digital-preservation/droid/issues/204


jcharlet:
very handy, thanks for that one.

jcharlet:
tested and verified on ubuntu

jcharlet:
(forgot to add a formal review)

Issue #329, opened 2019-12-01
DROID objects extremely hard to instantiate


nishihatapalmer:
There are a lot of DROID objects which are impossible to instantiate with Java, unless you use a dependency injection system like Spring.  They have private members and no programmatic way to set them.

Other classes can be instantiated, but then require multiple setters running to ensure that the object is in a safe state.  There is no indication what needs to be set on these classes, and what's optional.

It would be good if DROID objects could be instantiated using systems like Spring and also in a normal programmatic way via a constructor.

nishihatapalmer:
This has no effect on users of course; it just makes working with DROID code unnecessarily hard, and ties the code to Spring.

nishihatapalmer:
I am working on a PR which adds both empty constructors (for Spring like systems), and parameterized constructors to allow the objects to be constructed normally.

There's a lot of them, but the changes are minimal (adding two constructors to each affected object).  It should have no effect on the current running of DROID, because we are only adding constructors that nothing currently calls.

adamretter:
In general the data objects should be designed as POJOs (i.e. Java Beans), in this manner they can be instantiated either with or without Spring.

nishihatapalmer:
Unfortunately, quite a few of them aren't POJOs at all (no way to instantiate without Spring).  

The other ones which require multiple setters calling to ensure the object is safe to use are technically POJOs, but are hard to use safely.

PR #330, opened 2019-12-01
Constructors

nishihatapalmer:
There are a number of DROID objects which cannot be constructed safely without using a system like Spring.  They have private members, and no setters or constructors to set them.

There are also a number of DROID objects which require setting various internal dependencies to operate safely, but there are no constructors for this.  Instead, a programmer has to figure out what is needed for the class to operate safely, and to call the relevant setters.

This PR provides empty constructors (so that bean construction works as normal), and parameterized constructors, so that objects can be safely constructed programmatically.


nishihatapalmer:
This PR addresses issue: https://github.com/digital-preservation/droid/issues/329

nishihatapalmer:
So, there's a lot of changes in this PR, but they are almost all just adding an empty constructor and a parameterised constructor.

In some cases, without these constructors it is impossible to instantiate the objects (unless using a dependency injection framework like Spring), as they had private members which had no constructor or setter.

In other cases, it was possible to instantiate the object in an unsafe state (and then call lots of setters to make it safe).  Constructors are provided for these too, and they at least form documentation on what is required to safely instantiate one of these objects (even if no code currently calls it).

PR #333, opened 2019-12-06
add: comment in pom about hotfix on org-netbeans-swing-outline

jcharlet:
to ensure good maintenance and prevent upgrading library while not taking into account overridden class


jcharlet:
as discussed with @LauraDamianTNA

PR #335, opened 2019-12-06
fix: checkbox filter icon display on main screen

jcharlet:
fix display of filter checkbox on main screen

previously:
Screenshot from 2019-12-06 14-03-15 - 2

after fix:
Screenshot from 2019-12-06 15-15-53


LauraDamianTNA:
Perfectly aligned

PR #337, opened 2019-12-09
Shrink dependencies

nishihatapalmer:
There's a couple of unnecessary dependencies in DROID.

1.  spring transaction isn't actually used in DROID any more.
2.  aspect oriented programming (AOP) was being used to inject code into a single method.

This PR removes the references to unused dependencies, and gets rid of the completely unnecessary use of AOP, and instead just injects the required class in the single place it's needed using normal Spring.

### DIFF
--- a/droid-core-interfaces/pom.xml
+++ b/droid-core-interfaces/pom.xml
@@ -89,10 +89,6 @@
             <artifactId>jwat-common</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.aspectj</groupId>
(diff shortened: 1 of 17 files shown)


nishihatapalmer:
I have an ulterior motive for this PR.  The latest AOP rewriting breaks my somewhat old-but-still-good profiler, and I can't afford to update to the newer version.

Since the AOP is entirely unnecessary and just adds complexity for no good reason, I'd like to remove it in any case.

jcharlet:
rebased master and fixed conflicts

PR #338, opened 2019-12-13
Only parse container sigs once

nishihatapalmer:
At present the container signature file is parsed twice: once when the ZipIdentifier is instantiated, and again when the OLE2Identifier is.

This PR creates a new bean that caches the container signature definitions, and modifies the container identifiers to use the bean instead of parsing the file themselves.


jcharlet:
approved

jcharlet:
thanks for the work Matt!

Issue #166, opened 2017-06-23
Image scaling issues on high resolution screens or Surface Pro specific?


Dclipsham:
I've had a report of a user (Ed Pinsent) on a Win10 Surface Pro tablet that DROID hits image scaling issues. We need to understand whether this is specific to Surface tablets or generally with high res screens and DPI scaling, and deal with accordingly.
Representative screenshots below, used with permission
![appscaling](https://user-images.githubusercontent.com/2189778/27482022-9048664a-5817-11e7-99c3-93c2c1046a90.png)
![appscaling2](https://user-images.githubusercontent.com/2189778/27482023-9057d166-5817-11e7-8fe2-4181d0e9953b.png)

alexgreenDP:
This issue has been added to the DROID backlog, it will be assigned an appropriate priority but is unlikely to be included in the next release.

alexgreenDP:
Closed in error.

jcharlet:
@Dclipsham any idea of the resolution? Can't reproduce on 1920 * 1080

Dclipsham:
Afraid not - we weren't able to reproduce. We did have a second report from a user on a 'windows 10 laptop' but they were unable to provide us with further information re model, resolution etc. As unable to reproduce happy for this to pause and reopen if we get further reports.

jcharlet:
@Dclipsham could I get a surface pro from support to test on? 
As I said on trello, I have another display issue when building with java11 (but it's not on the exact same part of the interface), I'll be already investigating this one, maybe there is sthing related.

jcharlet:
Fixed in PR https://github.com/digital-preservation/droid/pull/289

jcharlet:
the following 2 PRs improve the icons which were pixelated when UI is scaled.

PR #344, opened 2019-12-30
Fix DROID GUI on high resolution screens with windows scaled

jcharlet:
Replace all icons with Java9 feature BaseMultiResolutionImage.

Basically, the previous PR, replaced all low quality icons with higher quality icons and scale them down depending on the need in the UI.
This PR provides icons in multiple formats (16x16 32x32 64x64 etc) and the app returns the best icon depending on current screen resolution. 

To enable building on Java8 on Travis, skip droid-swing-ui module when in Java 8, include it otherwise.

to fix https://github.com/digital-preservation/droid/issues/166 with https://github.com/digital-preservation/droid/pull/345
follows PR https://github.com/digital-preservation/droid/pull/345


MancunianSam:
This is 6 years old now and the issue attached to this is resolved. I'll close this. If the issue happens again we can open a new PR

PR #352, opened 2020-01-13
upd: migrate log4j to log4j2 to fix vulnerability on log4j 1

jcharlet:
update dependencies and configuration for logging to log4j2
to fix new vulnerability on log4j 1 which is preventing builds to succeed on CI.


jcharlet:
Hi all, if someone could review that PR, that would fix the vulnerability issue we have on all other builds, and help me move on. Thanks!

adamretter:
Hi @jcharlet there seems to be a huge amount of changes in this due to reformatting or something. Unfortunately for many but not all files in this PR that makes it impossible to pick out the intentional changes. 

If its not a huge amount of work, is it possible to get a PR where only the necessary code changes are made? or the reformatting changes are isolated in their own commit separately to the desired logging changes.

jcharlet:
> Hi @jcharlet there seems to be a huge amount of changes in this due to reformatting or something. Unfortunately for many but not all files in this PR that makes it impossible to pick out the intentional changes.
> 
> If its not a huge amount of work, is it possible to get a PR where only the necessary code changes are made? or the reformatting changes are isolated in their own commit separately to the desired logging changes.

Hi @adamretter , sure thanks, will do

jcharlet:
argh @adamretter there was no reformatting issue, it's because I removed log4j.properties files and created log4j2.properties files. And as it considers them as different files, it doesn't try to show the diff between them.

I updated the commits, so that I initially commit properties files with renaming them.
But changes are so big, github shows files one after another.

A bit easier to compare in intellij though..
Screenshot from 2020-01-16 11-31-13

Could that be satisfying? I agree it's a large change, all properties files were updated, and syntax changed from log4j 1 to log4j2..

jcharlet:
@adamretter I'll save you some time, we have a new developer Saurab who's available to review the PR :), will see that with him!

droid-binary/conf/log4j2.properties
```diff
@@ -1,81 +1,81 @@
-#
-# Copyright (c) 2016, The National Archives <pronom@nationalarchives.gsi.gov.uk>
```
sparkhi: The email id should be corrected. It is not strictly part of this PR, so it should be addressed on its own with corrections in all places where it appears in the codebase, also look to see if it forces a copyright year update

sparkhi:
The email id in the source header should be corrected. It is not strictly part of this issue, so it should be addressed on its own with corrections in all places where it appears in the codebase, also look to see if it forces a copyright year update

Issue #365, opened 2020-01-29
manage warc 1.1


sparkhi:
### Context
Droid does not open WARC 1.1 format files. It needs to open that. As part of the work, there will be 2 formats for 1.0 and 1.1 and any older files will be opened by the existing format 

### Changes 

- Get new format in the signature file 
- Include the new formats in the properties file for UI, CLI and interfaces 
- Make changes in the CLI to be able to pick all formats of WARC

### Acceptance Tests GUI

-   Get test data in different WARC formats (specifically 1.0, 1.1 and an older one such as 0.17) 
- Add these files to a profile and click on the 'start' button
- All 1.0 and 1.1 files should be correctly identified by their version number, 0.17 will show the version number as blank 
- All files should be expanded to show the hierarchy 

### Acceptance Tests CLI

- Launch droid command line with parameters to analyse the test data files 
- On command line all of them should be successfully identified 
- On command line the output should list its hierarchical structure using slashes for folder separation

NOTE: The new signature file is not yet available on release so testing will need to be done using a local updated signature file

PR #367, opened 2020-01-29
Manage opening of WARC 1.1 and update identification of WARC 1.0

sparkhi:
Updates to support Warc 1.1, In addition to that, there are changes to 
1) Support newly created signature for WARC1.1 (fmt/1355) which is separate from generic WARC 
2) Fixed an NPE in CLI which happened when the date string in WARC included fractions of second
3) Added a couple of developers names in parent POM

[UPDATE Jeremie] Fix https://github.com/digital-preservation/droid/issues/365


jcharlet:
@sparkhi can you please provide test files on the repo and link to them here? 
Should we add a test case for each test file?

sparkhi:
The CI failure on Travis has a miniscule chance of happening when the timestamp for the file has milliseconds part as zero. I'll fix that as well.

sparkhi:
[testing-365.zip](https://github.com/digital-preservation/droid/files/4134443/testing-365.zip)
The zip contains test data. There is at least one file each which will identify against each of the 3 puids for warc. There are a couple other files as well. One error file to make sure there is no bad effect when a file is not identified

sparkhi:
[devtest-467.zip](https://github.com/digital-preservation/droid/files/4134587/devtest-467.zip)
The signature file includes new signature for 1.0 - fmt/1355

Issue #362, opened 2020-01-27
DROID opens multiple connection pools to the same database.


nishihatapalmer:
When DROID starts, it opens multiple connection pools to the same database.  This can be seen in the logs when it starts, the `HikariPool` is initialised multiple times.  This is slow and unnecessary.

I've traced why this happens in the git history.  

1.  DROID used to have a single data source for the embedded derby database (which is all that is needed).  
2.  A connection pool was added to improve performance.  It's unlikely that this improved performance, since DROID opens a few connections to the database and holds on to them as long as it needs to.  It isn't a server application trying to share scarce database connections among thousands of concurrent connections.  
3.  A factory for the connection pool was added later.  This seems to be because of strange initialisation / dependency issues in spring initialisation.   The factory also reloads the driver class every time it creates a new pool.
4.  All the things that used to get the connection pool now get a new connection pool from the factory.

The net result is that all the things that just need an occasional database connection now get a connection pool to themselves (and try to reload the database driver).

nishihatapalmer:
I'm working on how to fix this.  Everything still works, so there's no massive issue here, but it would be nice to clean this up.

nishihatapalmer:
Main classes are `DerbyPooledDataSource` and `DerbyPooledDataSourceFactory`.

adamretter:
@nishihatapalmer It's probably a Spring scope thing. If we set the Spring scope to `singleton` for the `DerbyPooledDataSourceFactory` we should only get one per JVM instance

nishihatapalmer:
@adamretter  I think there's only one factory, but it creates a new DerbyPooledDataSource for each object that wants a data source.

jcharlet:
brilliant, thanks Matt for looking into this!

nishihatapalmer:
I've created a PR that fixes this issue (in that we only create one connection pool per profile).

See the PR for some additional comments.

rhubner:
Hi all my friends,

I think it can be fixed here : 
https://github.com/digital-preservation/droid/blob/master/droid-results/src/main/java/uk/gov/nationalarchives/droid/profile/datasource/DerbyPooledDataSourceFactory.java#L141

Mark it as singleton. It should create only one DB pool per application context. Not as @adamretter mentioned per JVM. I'm not sure at the moment how this work when you have hierarchy of application contexts. Each should be separate application context, but they should inherit from master application context.

nishihatapalmer:
Yep, that works :)

Many thanks Radek!

[...]

PR #368, opened 2020-01-29
Dbconnections

nishihatapalmer:
Removes the factory which created a new connection pool for everything that just wanted a connection pool.

See issue https://github.com/digital-preservation/droid/issues/362

Now we just use the connection pool to give classes connections, rather than creating a new pool for each class.

Needs some testing, as the reason for using the factory in the first place is not defined.  There were some comments about fixing some spring initialization error, but there are no recorded issues in GitHub about this.

Regardless, using a factory to create a new connection pool is not the solution to initialization issues in the connection pool, as we don't want separate connection pools for each class.


nishihatapalmer:
So, merging latest master to resolve the conflicts was pretty easy, but now I get weird spring initialization problems.  

The data source is now being initialized with a null HikariConfig object, even though nothing in the spring-jpa has changed.  I guess this was why the factory was erroneously introduced - to try to force spring to initialize objects in the right order.

Something is clearly weird here.  No change to spring configuration but changes to the order of initialization.

nishihatapalmer:
OK - @rhubner has provided a better solution - to set the DerbyPooledDataSourceFactory isSingleton() method to true.  Which apparently ensures the factory only creates a single object (I thought it meant only one factory was produced...)

jcharlet:
thanks Matt for looking into this, too bad Radek provided a solution already, hope you didn't waste too much time!

jcharlet:
sorry too quick to press buttons, I suppose you might want to keep your work aside

PR #369, opened 2020-01-29
Ensure only a single object is created.

nishihatapalmer:
Ensures the factory only creates a single pool per application context.


nishihatapalmer:
Fixes issue: https://github.com/digital-preservation/droid/issues/362

Issue #360, opened 2020-01-23
CSV export incorrect


jcharlet:
from @dclipsham: one thing i picked up on with the build @Jeremie gave to me - when exporting the CSV, the format identifier columns were not correct - no PUID, name or version listed. I gotta leave now but can replicate tomorrow if needed

PR #372, opened 2020-01-31
fix: format version and name was not given anymore in exports

jcharlet:
# issue
from @dclipsham: one thing i picked up on with the build @Jeremie gave to me - when exporting the CSV, the format identifier columns were not correct - no PUID, name or version listed. I gotta leave now but can replicate tomorrow if needed
https://github.com/digital-preservation/droid/issues/360

# context
introduced on https://github.com/digital-preservation/droid/pull/212 745cc37d0bfcaf5f0f3229bf5a6c2b8e34bb59fa

# solution
put back previous code where we retrieve name and version from jdbc query (rather than constant)

# Acceptance test
- run analysis of a folder on GUI
- make an export
- check that version and name columns are now provided in the export

### DIFF
--- a/droid-results/pom.xml
+++ b/droid-results/pom.xml
@@ -302,10 +302,6 @@
 			<artifactId>HikariCP-java7</artifactId>
 			<version>2.4.13</version>
 		</dependency>
-		        <dependency>
-            <groupId>org.apache.logging.log4j</groupId>
-            <artifactId>log4j-slf4j-impl</artifactId>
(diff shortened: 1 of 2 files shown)


jcharlet:
[export_broken.txt](https://github.com/digital-preservation/droid/files/4140699/export_broken.txt)
[export_fixed.txt](https://github.com/digital-preservation/droid/files/4140700/export_fixed.txt)

PR #373, opened 2020-01-31
Only log hikari errors.

nishihatapalmer:
* removes noise from the main DROID log in normal operation.

Note: you won't see these changes unless you get rid of your .droid6 folder.  If you have an existing installation, DROID will continue to use the previous log4j2.properties file already installed.

### DIFF
--- a/droid-binary/conf/log4j2.properties
+++ b/droid-binary/conf/log4j2.properties
@@ -77,5 +77,6 @@ log4j.category.org.springframework=ERROR
 log4j.category.org.hibernate=ERROR
 log4j.logger.org.springframework=ERROR
 log4j.logger.org.hibernate=ERROR
+log4j.category.com.zaxxer.hikari=ERROR
 
 
--- a/droid-swing-ui/src/main/resources/log4j2.properties
+++ b/droid-swing-ui/src/main/resources/log4j2.properties
@@ -70,5 +70,8 @@ logger.apache.level=error
 logger.apache.appenderRef.console.ref = console
 logger.apache.appenderRef.rolling.ref = rolling
 
-
+logger.hikari.name=com.zaxxer.hikari
+logger.hikari.level=error
+logger.hikari.appenderRef.console.ref = console
+logger.hikari.appenderRef.rolling.ref = rolling
 


sparkhi:
Hi @nishihatapalmer, 
Could you please post some instructions which would help me test it. It would be great to see the problem and then test after applying the changes you have provided.

TIA,

nishihatapalmer:
**Problem**
HikariPool log messages appear in the main DROID log.   They are harmless, but are noise.   Each time a profile is created in the GUI version of DROID, you will see a HikariPool initialising itself in the log output.

**Solution**
Ensure log messages from Hikari are only at ERROR level.  This is done by altering the log4j properties file.  You will be able to see that we have similarly disabled log output messages for apache and other libraries we use.

**Testing**
I'm not aware of a way to manually test this, although there is probably some magic that would let you do it.   The easiest way to test is to run DROID and observe the log output.  You should see HikariPool messages.  Then, delete or rename the .droid6 folder under your home directory, and re-run DROID  This will force DROID to recreate it, putting in the new log4j property file there.  Each time a profile is created, you should no longer see HikariPool messages in the DROID log.

You have to ensure that the new resources are used by DROID.  When DROID is installed, it creates a .droid6 folder under the user's home directory, and puts things like the log4jconfiguration there.

sparkhi:
- I checked out the revision corresponding to 'prepare release droid-6.4'. 
- I deleted the .droid6 folder
- Built locally 
- Ran the local version 
I did not see any specific message related to hikaripool, 

Do you reckon I've gone too far back? I checked the console as well as the log file under .droid6/logs and could not find the messages. Could you point me in the right direction?

nishihatapalmer:
I don't know when this issue surfaced, but it's in 6.5-SNAPSHOT.

You need a log4j properties file from any old version.  If you've run 6.4
then you will have the old log4j file.

If you run the latest master, you should see HikariPool log entries.

If you apply this PR (which bundles an updated log4j file), you should
still see the HikariPool entries (because you are still using the old log4j
file in the .droid6 folder).

If you then rename or delete the .droid6 folder and re-run DROID with this
PR, the HikariPool entries should be gone (because DROID has recreated the
.droid6 folder and put the updated log4j property file in it).

nishihatapalmer:
Essentially, all this PR does is to ensure log entries from Hikari only appear at an ERROR level,  the same as we do for apache and some others.

To test, you need the log4j property file from the last version of DROID, and the one from this PR.

If you run the 6.5-SNAPSHOT master of DROID with a standard log4j property file in the .droid6 folder, you should see HikariPool entries.

If you replace the log4j property file with the one from this PR, then the HikariPool entries should disappear.

nishihatapalmer:
It does raise the issue of what we do with things like log configuration updates to DROID.  Anyone who's had any previous version installed simply won't get these updates, unless they blow away their .droid6 folder.

Should DROID check for differences and put in the new version if it's not there with a warning message?  Or rename the old file and put the new one in?

PR #374, opened 2020-02-01
Identify containers first

nishihatapalmer:
Separates core/container signature identification from results processing, and uses a slightly more efficient matching strategy (which produces identical results).

**Separate identification from results processing.**

DROID currently splits signature identification between the BinarySignatureIdentifier which handles binary signatures and file extensions, and the SubmissionGateway which does the container signature processing.  This is not "by design" - container signature matching was just shoe-horned into the SubmissionGateway.

Fundamentally, the SubmissionGateway should not know how signatures are identified.  It should run files through identification, schedule threads, store the results, and process any archive formats it has recognised.

Logically separating this functionality allows different matching strategies to be explored and compared, and puts all the identification code in one place.

**More efficient matching strategy**
DROID currently follows this strategy for matching files:

1.  Run all binary signatures on a file (BinarySignatureIdentifier)
2.  Run any container signatures if we have a match for the base type (e.g. OLE2).
3.  If we have a container signature match, just use that and discard any previous binary matches.
4.  If we don't have a container match, use the binary matches we found first.

Note that if we have a match for a container signature, we discard all the binary matches we may have found.  We can call this strategy BinaryAndContainerMatching.

The new strategy is to look for container matches first (just OLE2 and ZIP binary sigs and the corresponding container signatures).  If we find a container match, we take the match.  If we don't find a container match, we run the binary signatures and take those results.  We can call this strategy ContainerOrBinaryMatching.

This gives exactly the same matches as the first strategy, but avoids running all the binary signatures if we have a container signature match.  In practice, it is a little faster (less than 5%) - most of DROID's processing time is actually taken up with file IO, database writes and archive file processing, not signature matching.


nishihatapalmer:
Although there appear to be a lot of changes here, most of them are either merging in changes from master, or moving the identification tests and files from the core package to the container package.

**Moving identification to container package.**

The existing BinarySignatureIdentifier was defined in the droid-core package.  However, we don't want to add container signature processing into there, as it would create a cyclic dependency between core and container.  So, we move the BinarySignatureIdentifier to the droid-container package, and rename it SignatureIdentifier.

Along the way, we amend the DroidCore interface to support container matching, and also add a general purpose match() method, which will run whatever match strategy we want.  

A consequence of moving the identifier to the container package is that all the tests and files associated with it have to move over from core to container too - so that accounts for quite a large number of the changes.

All the identification code (and associated parameters such as maxBytesToScan) are also moved from SubmissionGateway to the SignatureIdentifier.  These are now fully separate, and different identification strategies can be plugged in to the SubmissionGateway - it doesn't care how it's done (which is as it should be).

nishihatapalmer:
**Testing**

I have run multiple comparisons between this code and the previous, and results are identical between them.  Loading a profile from master and one created by this branch, and running DROID's own reports across both profiles shows that all formats are identified exactly the same.

Different parameters for maxBytesToScan, and whether archive files are processed or not were also tried, with identical results in all cases.

nishihatapalmer:
**Note**: this branch started as an attempt to explore different matching strategies - specifically the container first strategy.  However, this was basically impossible with the current split between identification and result processing, which led to rationalising where the identification code lived, and taking it out of the SubmissionGateway.  

It's evident that this split was already a concern if you look at the comments in DroidCore:

```
/**
 * DROID core API.
 * Call submit() to submit identification requests.
 * 
 * This interface is somewhat artificial, reflecting the original split
 * between binary signature processing and later forms of identification
 * (container signatures).  Since these signatures interact, we require
 * "fix-up" methods here.
 * 
 * Specifically:
 * 
 * 1) If there is a container signature for a puid, we do not
 * want to run any defined binary signatures for the same puid, as this 
 * would be redundant.
 * 
 * 2) All file formats and their extensions are defined in the binary
 * signature file alone (so exist in implementations of this interface).
 * However, we need to be able to perform extension checking and 
 * extension mismatch checking on container signature results
 * as well as for binary signature results, so there are public methods
 * to allow this.
 * 
 * Eventually, it would be a good idea to have a single identification interface,
 * with separate interfaces / classes to perform these other individual functions
 * accessible behind the main identification interface.  
 * 
 * @author rflitcroft
 *
 */
```

nishihatapalmer:
**Command line differences**

The container first strategy will only be applied in the UI, not the command line project.  This is because the command line project implements its own container processing (reflecting the original split between the submission gateway and the BinarySignatureIdentifier).  So it will continue to do binary signature identification first, followed by container signature processing.

The new SignatureIdentifier is backwards compatible with the BinarySignatureIdentifier, in that it can still do standard binary signature processing.  The original interfaces were completely preserved, so the command line should continue to function exactly as before.

jcharlet:
Matt, thanks again for the work, this is quite a significant change, so we'll integrate it after we released what we currently have.

nishihatapalmer:
I agree, it is a fairly significant change.  I'm also working on applying this to the command line project, which I think is part of this change.

nishihatapalmer:
This PR is massively out of date now, given the PR recently approved to the command line project to remove duplicate code.
I will resubmit in a new PR.

Issue #371, opened 2020-01-30
GUI gets stuck forever while trying to browse sub archives in fake archives from skeleton suite


jcharlet:
# Issue
fake archives created in skeleton suite makes the GUI fall into an infinite loop while trying to process fake archives from skeleton suite

Processing works if we don't expand archives

still to test on CLI

# Context
another collateral damage of PR https://github.com/digital-preservation/droid/pull/334
depends on https://github.com/digital-preservation/droid/pull/376

# Solution
We probably need to now manage corrupted archives

jcharlet:
another possibly related issue: CLI on skeleton suite throws NPE on Bzip2 embedded archive

jcharlet:
works fine when we disable option to expand archives and web archives

PR #376, opened 2020-02-03
fix: CLI can browse again folders recursively

jcharlet:
# Issue
Tried using the -R flag to recurse folders ./droid.sh -Nr "/home/saurabh/test-data/warc" -Ns "DROID_SignatureFile_V96.xml" -Wt WARC -R
---- It did not recurse the folders underneath which also had WARC files

fix https://github.com/digital-preservation/droid/issues/363

# Context
bug introduced in droid/droid-command-line/src/main/java/uk/gov/nationalarchives/droid/command/action/NoProfileRunCommand.java on commit 0d3cbe2 from #167

# Acceptance tests
- run CLI on folder containing sub folders
with following arguments (adapt to your workspace)

```
-R
-Nr
/boot/grub
-Ns
/home/jeremie/.droid6/signature_files/DROID_SignatureFile_V91.xml
-Nc
/home/jeremie/.droid6/container_sigs/container-signature-20170330.xml
```

- verify CLI processed sub folders

PR #378, opened 2020-02-03
Fix skeleton issue 

jcharlet:
# Issue
since latest work on CLI and GUI on managing archives and web archives, we couldn't run the CLI or GUI on the skeleton suite without disabling expansion of sub-archives.

multiple failures on skeleton suite files (which are valid regarding signatures, but not actually valid files that you can open, so we can consider them as corrupted), due to multiple reasons:
- the difference in the management of different formats (some raise exceptions, some just log the error)
- null pointer exception when trying to close the reader on non-existing subfiles

# Context
fix https://github.com/digital-preservation/droid/issues/371
fix another bug previously identified: bug: CLI on skeleton suite throws NPE on Bzip2 embedded archive

depends on https://github.com/digital-preservation/droid/pull/376 because it makes testing simpler

# Solution
only log error when processing corrupted files
prevent NPE on reader closing

# Acceptance tests
### Run on fixed DROID and export results
- run CLI on skeleton folder in repository droid/droid-core/test-skeletons
- run GUI on skeleton folder
- compare exports CLI vs GUI and check there is no difference
### Run on Droid 6.4 and export results
- run Cli and export
- run GUI and export
- compare exports CLI vs GUI and check there is no difference
### Compare exports
- compare export GUI on Fixed DROID vs 6.4
- compare export CLI on Fixed DROID vs 6.4

### DIFF
--- a/droid-command-line/src/main/java/uk/gov/nationalarchives/droid/command/archive/ArchiveContentIdentifier.java
+++ b/droid-command-line/src/main/java/uk/gov/nationalarchives/droid/command/archive/ArchiveContentIdentifier.java
@@ -60,6 +60,10 @@ public abstract class ArchiveContentIdentifier {
     protected String path;
     private ArchiveConfiguration archiveConfiguration;
 
+    protected static final String START_PARENTHESIS = " (";
+    protected static final String END_PARENTHESIS = ")";
+
+
     /**
      * Initialization of instance values must be explicitly called by all children.
      * @param binarySignatureIdentifier     binary signature identifier
--- a/droid-command-line/src/main/java/uk/gov/nationalarchives/droid/command/archive/Bzip2ArchiveContentIdentifier.java
+++ b/droid-command-line/src/main/java/uk/gov/nationalarchives/droid/command/archive/Bzip2ArchiveContentIdentifier.java
@@ -57,9 +57,6 @@ public class Bzip2ArchiveContentIdentifier extends ArchiveContentIdentifier {
 
     private static final long SIZE = 12L;
     private static final long TIME = 13L;
-    private static final String START_PARENTHESIS = " (";
-    private static final String END_PARENTHESIS = ")";
-
 
     /**
      *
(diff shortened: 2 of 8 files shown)


sparkhi:
J: regarding acceptance tests use the same signature files for both 6.4 and 6.5
Maybe sort output files using `sort` linux CLI before comparing

jcharlet:
I think the use of sys.err is still relevant, because we want to be able to
log stuff back to the user, while disabling app logging entirely. I would
leave it as it is.
ᐧ

sparkhi:
Happy with the CLI output, More formats are recognised, one is correctly recognised as compared to 6.4. Now going through the GUI results

sparkhi:
Tested between 6.4 and 6.5, The results are expected. The places where there is a difference between 6.4 and 6.5 is listed below. please note that these differences are expected. 
### GUI
Files that were identified as 'File' in 6.4, but identified as 'Container' in 6.5
	fmt-289-signature-id-305.warc
	fmt-410-signature-id-580.arc
	x-fmt-219-signature-id-525.arc

Files that were identified as 'File' in 6.4 and status 'Done', but identified as 'Container' in 6.5 and status 'Error'
	x-fmt-267-signature-id-340.bz

### CLI
The results are identical

### Console output when running GUI
Same error in both versions:
--- No reader was obtained for file:/test-data/test-skeletons/fmt/fmt-40-signature-id-182.doc 
--- No reader was obtained for file:/test-data/test-skeletons/fmt/fmt-125-signature-id-171.ppt
--- Could not process the archival format(ISO): file:/test-data/test-skeletons/fmt/fmt-468-signature-id-730.iso	java.io.IOException: Could not deserialize a complete block
--- No reader was obtained for file:/test-data/test-skeletons/fmt/fmt-126-signature-id-172.ppt
--- Could not process the archival format(7Z): file:/test-data/test-skeletons/fmt/fmt-484-signature-id-756.7z
--- No reader was obtained for file:/test-data/test-skeletons/fmt/fmt-39-signature-id-128.doc
--- No reader was obtained for file:/test-data/test-skeletons/fmt/fmt-111-signature-id-170.nul
--- Could not process the archival format(BZ): file:/test-data/test-skeletons/x-fmt/x-fmt-268-signature-id-342.bz2	BZip2 block size is invalid
--- No reader was obtained for file:/test-data/test-skeletons/x-fmt/x-fmt-88-signature-id-181.ppt.
--- Could not process the archival format(GZ): file:/test-data/test-skeletons/x-fmt/x-fmt-266-signature-id-201.gz	null

Error only in 6.4:
--- No reader was obtained for file:/test-data/test-skeletons/x-fmt/x-fmt-430-signature-id-208.msg.

Error only in 6.5
--- Could not process the archival format(BZ): file:/test-data/test-skeletons/x-fmt/x-fmt-267-signature-id-340.bz	Stream is not in the BZip2 format

Issue #395, opened 2020-02-21
cannot open folder shortcuts on Windows


jcharlet:
# Issue
reported by @paulyoung84 on Windows 10 using JRE bundle version (Java 11)

> on first test (just loading a profile) it works fine for most folders, but selecting a folder from top level, like through libraries or folder on my desktop it seems to struggle. tries to load but files don't appear in window.

# Further investigation 
as observed with @sparkhi , seems to happen on shortcuts of network folders, typically "Libraries" or "Desktop" when taken from the top level in file hierarchy. But when you browse to the right drive without using the library, e.g. H:/users/username/Desktop, it works fine.

on 6.4, it does work but the name of folder is weird: "::{12342134}" instead of "Libraries".

on 6.5, it does not work indeed, and the following stacktrace appears in the logs:

```
Exception in thread "AWT-EventQueue-0" java.nio.file.InvalidPathException: Illegal char <:> at index 0: ::{031E4825-7B94-4DC3-B131-E946B44C8DD5}
        at java.base/sun.nio.fs.WindowsPathParser.normalize(Unknown Source)
        at java.base/sun.nio.fs.WindowsPathParser.parse(Unknown Source)
        at java.base/sun.nio.fs.WindowsPathParser.parse(Unknown Source)
        at java.base/sun.nio.fs.WindowsPath.parse(Unknown Source)
        at java.base/sun.nio.fs.WindowsFileSystem.getPath(Unknown Source)
        at java.base/java.io.File.toPath(Unknown Source)
        at uk.gov.nationalarchives.droid.gui.action.AddFilesAndFoldersAction.add(AddFilesAndFoldersAction.java:92)
        at uk.gov.nationalarchives.droid.gui.DroidMainFrame.jButtonAddFileActionPerformed(DroidMainFrame.java:1367)
        at uk.gov.nationalarchives.droid.gui.DroidMainFrame$10.actionPerformed(DroidMainFrame.java:543)
        at java.desktop/javax.swing.AbstractButton.fireActionPerformed(Unknown Source)
        at java.desktop/javax.swing.AbstractButton$Handler.actionPerformed(Unknown Source)
        at java.desktop/javax.swing.DefaultButtonModel.fireActionPerformed(Unknown Source)
        at java.desktop/javax.swing.DefaultButtonModel.setPressed(Unknown Source)
        at java.desktop/javax.swing.plaf.basic.BasicButtonListener.mouseReleased(Unknown Source)
[...]

PR #404, opened 2020-02-27
Manage special folders (forbidden, shortcuts)

sparkhi:
[Update Jeremie]

# Issues
https://github.com/digital-preservation/droid/issues/395
https://github.com/digital-preservation/droid/issues/406

# Solution
- Added a validation to make sure non - fs folders like 'Libraries' onwindows cannot be added
- manage forbidden files and folders: prevent breaking + show as forbidden in GUI

# Acceptance Test
NOTE: The testing can only be done on Windows OS for this fix, other OSs should not see any change.

1. Open Droid GUI
2. Create a new profile
3. Click on the 'add' button to bring up file chooser dialog
4. Pick one of the Windows 'virtual collection' folder (e.g. Libraries, Documents) which does not map directly to file system
5. You should see an error message informing you that the folder is not allowed.
6. If there is more than one non-file-system folder picked in step #4 above, all such folder names should appear in the error
7. If there is a combination of non-file-system folder and a file-system-folder, the error message should inform only about the non file-system folder
8. If all folders picked in step #4 are valid file-system folders, they should all be added to the profiling dialog.


sparkhi:
### Acceptance Test

NOTE: The testing can only be done on Windows OS for this fix, other OSs should not see any change.

1. - Open Droid GUI
2. - Create a new profile
3. - Click on the 'add' button to bring up file chooser dialog
4. - Pick one of the Windows 'virtual collection' folder (e.g. Libraries, Documents) which does not map directly to file system
5. - You should see an error message informing you that the folder is not allowed. 
6. - If there is more than one non-file-system folder picked in step #4 above, all such folder names should appear in the error 
7 - If there is a combination of non-file-system folder and a file-system-folder, the error message should inform only about the non file-system folder 
8 - If all folders picked in step #4 are valid file-system folders, they should all be added to the profiling dialog.

jcharlet:
fix https://github.com/digital-preservation/droid/issues/395

jcharlet:
works on libraries and similar folders, but crashes on forbidden folders

jcharlet:
as discussed with Saurab, this PR will fix as well https://github.com/digital-preservation/droid/issues/406

Issue #215, opened 2018-07-26
Ability to get and set maxBytes param using DROID CLI


Dclipsham:
Request via Google Group - https://groups.google.com/forum/#!topic/droid-list/YiV7NQE2olk

nishihatapalmer:
This is really wider than just max bytes to scan.  We really need to be able to pass any configurable profile property into the CLI.  

At present, we have to edit the profile properties in the GUI (not workable if you want to script different things) or use scripting hacks to edit the properties file just to control what the CLI does.

nishihatapalmer:
A simple method might be to offer a command line option that allows particular properties to be set, as some kind of delimited string.  It might be a bit geeky vs. having dedicated command line options for each setting,, but it would provide a default capability for any configurable property.

nishihatapalmer:
I'm having the issue of injecting runtime configuration into a new profile in some other work I'm doing.

When DROID initialises a profile, it reads configurable properties from the global config from the properties file in the profile working directory.  

It's not clear how to inject additional property overrides (or to specify another property file) given the way it's all initialized by Spring.  I guess there must be a way to inject runtime properties into the spring initialization...

nishihatapalmer:
Right - I think the ProfileManager interface needs a new create method which takes sigs AND a set of properties to override any default profile properties.  The current method only lets you create a profile with sigs passed in (so all runtime configuration comes from default properties).

nishihatapalmer:
I have a branch on my own fork which allows any profile properties to be set, either directly on the command line, or by specifying an alternate property file to use.

Still needs more testing, but can be seen here: https://github.com/nishihatapalmer/droid/tree/ProfileProperties

To override a profile property such as max bytes to scan, you specify it like this:

`droid -a "/home/user/Documents" -R -p "/home/user/docprofile.droid" -Pr "profile.maxBytesToScan=1024"`

If you have a lot of properties you want to override all the time, or different sets of properties you want to re-use, you can put them in a property file and tell DROID to use that:

`droid -a "/home/user/Documents" -R -p "/home/user/docprofile.droid" -Pf  "/home/user/my.properties"`

nishihatapalmer:
One slightly annoying limitation of the current code is that a few properties are not set when a profile is created, but are rather used when it is opened (and taken from the global property file at that point).

[...]

PR #407, opened 2020-03-02
Profile properties

nishihatapalmer:
Allows global default profile properties to be overridden when creating a new profile.  This allows us to set profile properties directly on the command line, or by specifying an alternate profile property override file to use, or both.

**Usage**
To override a default profile property such as max bytes to scan, you specify it with the **-Pr** option:

`droid -a "/home/user/Documents" -R -p "/home/user/docprofile.droid" -Pr "profile.maxBytesToScan=1024"`

You can set more than one property by just specifying each as additional arguments:

`droid -a "/home/user/Documents" -R -p "/home/user/docprofile.droid" -Pr "profile.maxBytesToScan=1024" "profile.processZip=false"`

If you have a lot of properties you want to override all the time, or different sets of properties you want to re-use, you can put them in a property file and tell DROID to use that with the **-Pf** option:

`droid -a "/home/user/Documents" -R -p "/home/user/docprofile.droid" -Pf "/home/user/my.properties"`

You can also specify properties on the command line with -Pr and use a property file with -Pf. In this case, any properties specified directly on the command line take precedence over the ones in the file (which in turn has precedence over properties defined in the droid global property file).

`droid -a "/home/user/Documents" -R -p "/home/user/docprofile.droid" -Pr "profile.maxBytesToScan=1024" "profile.processZip=false" -Pf "/home/user/my.properties"`

To see all the profile properties that can be set, look at the droid.properties file in the droid home folder. This is normally located in the .droid6 folder under the user's home folder.

### DIFF
--- a/droid-command-line/src/main/java/uk/gov/nationalarchives/droid/command/action/CommandFactoryImpl.java
+++ b/droid-command-line/src/main/java/uk/gov/nationalarchives/droid/command/action/CommandFactoryImpl.java
@@ -34,7 +34,10 @@ package uk.gov.nationalarchives.droid.command.action;
 import java.io.PrintWriter;
 
 import org.apache.commons.cli.CommandLine;
+import org.apache.commons.configuration.CombinedConfiguration;
+import org.apache.commons.configuration.ConfigurationException;
 import org.apache.commons.configuration.PropertiesConfiguration;
+import org.apache.commons.configuration.tree.OverrideCombiner;
 import org.apache.commons.lang.StringUtils;
 
 import uk.gov.nationalarchives.droid.command.FilterFieldCommand;
@@ -206,12 +209,40 @@ public class CommandFactoryImpl implements CommandFactory {
         command.setDestination(destination[0]);
         command.setResources(resources);
         command.setRecursive(cli.hasOption(CommandLineParam.RECURSIVE.toString()));
+        command.setProperties(getOverrideProperties(cli));
+        return command;
+    }
 
+    private PropertiesConfiguration getOverrideProperties(CommandLine cli) throws CommandLineSyntaxException {
+        PropertiesConfiguration overrideProperties = null;
+
+        // Get properties from a file:
+        final String propertyFile = cli.getOptionValue(CommandLineParam.PROPERTY_FILE.toString());
+        if (propertyFile != null && !propertyFile.isEmpty()) {
+            try {
+                overrideProperties = new PropertiesConfiguration(propertyFile);
+            } catch (ConfigurationException e) {
+                throw new CommandLineSyntaxException(e);
+            }
+        }
+
+        // Get properties from the command line directly:
         final String[] propertyOverrides = cli.getOptionValues(CommandLineParam.PROFILE_PROPERTY.toString());
         if (propertyOverrides != null && propertyOverrides.length > 0) {
-            command.setProperties(createProperties(propertyOverrides));
+            PropertiesConfiguration commandLineProperties = createProperties(propertyOverrides);
+            if (overrideProperties == null) {
+                overrideProperties = commandLineProperties;
+            } else {
+                CombinedConfiguration combined = new CombinedConfiguration();
+                combined.setNodeCombiner(new OverrideCombiner());
+                combined.addConfiguration(commandLineProperties);
+                combined.addConfiguration(overrideProperties);
+                PropertiesConfiguration merged = new PropertiesConfiguration();
+                merged.append(combined);
+                overrideProperties = merged;
+            }
         }
-        return command;
+        return overrideProperties;
     }
 
     @Override
--- a/droid-command-line/src/main/java/uk/gov/nationalarchives/droid/command/action/CommandFactoryImpl.java
+++ b/droid-command-line/src/main/java/uk/gov/nationalarchives/droid/command/action/CommandFactoryImpl.java
@@ -34,6 +34,7 @@ package uk.gov.nationalarchives.droid.command.action;
 import java.io.PrintWriter;
 
 import org.apache.commons.cli.CommandLine;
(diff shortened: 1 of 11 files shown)


nishihatapalmer:
Addresses the ability to set max bytes to scan raised in this issue:

https://github.com/digital-preservation/droid/issues/215

nishihatapalmer:
**Limitations**
One slightly annoying limitation of the current code is that a few properties are not used when a profile is created, and are not saved to the profile, but are rather obtained when it is opened (and taken from the global property file at that point).

All the normal profile properties like whether it processes archive files, or max bytes to scan can be set on creation. These properties tend to be prefixed by profile. But there are other properties defined in the global property file too, like database.durability. These will still be taken from the global property file when the profile is actually opened, so they can't currently be overridden.

To fix that, we'd need a profile to remember all the properties it is given when created (not just the ones that directly affect the profile creation), so that they could be re-used when it is opened. This is harder than it first seems - I've experimented with this and it runs into some tricky issues, for example, how persistent those profile settings are.

nishihatapalmer:
These are the profile properties which can be set when it is created:

```
profile.defaultThrottle 
profile.defaultBinarySigFileVersion 
profile.defaultContainerSigFileVersion 
profile.processTar 
profile.processZip
profile.processGzip 
profile.processRar 
profile.process7zip 
profile.processIso
profile.processBzip2
profile.processArc
profile.processWarc
profile.maxBytesToScan
profile.matchAllExtensions
profile.generateHash
profile.hashAlgorithm
```

jcharlet:
Brilliant, thanks Matt, will include that in next release (currently stopping PRs on the current release apart from hotfixes to pass the testing)

nishihatapalmer:
@jcharlet I thought I'd probably missed the 6.5 window.

I did this PR for somewhat selfish reasons, as I need to pass in profile properties for another PR I'm working on, which is to replace "no profile database mode" and all associated command line archival processing in the command line project.    So that will end up depending on this one.

No worries if this PR is rejected for some reason at a future date.  I'm sure most of the required code will be somewhat similar, even if it's not precisely the code here.

nishihatapalmer:
I'm now of the opinion that only global properties should be persisted, as per the current design.  The changes required to persist arbitrary properties are extensive, and are not particularly useful to be persisted.  Some you may want to persist per profile, others you may not want to (e.g. if debugging a database you may want to change the durability property after pausing and reloading to carry on).

There is only one use case where I want to set a property that relates to the profile itself, which isn't currently a global property, and that is a new one that allows the profile to output to a file instead of a database.  There is actually a good use case for this to be one of the main profile properties, as it tells DROID where to send output for that specific profile.  This is by default the database right now, but with the option of either database or file, a profile should know where its output should be directed to.

jcharlet:
created a separate backup branch
rebased master on matt's branch on matt's repo

jcharlet:
so the CI runs fine on linux and apple, not on windows on appveyor :/ 
[log (2).txt](https://github.com/digital-preservation/droid/files/4870078/log.2.txt)

jcharlet:
while the test runs just fine on my windows laptop :/

jcharlet:
can't reproduce locally,
new run of exact same code on appveyor shows the same test in error but at a different line. There must be a delay or something.

```
[00:04:04] [ERROR] Failures: 
[00:04:04] [ERROR]   ProfileManagerImplTest.testOverrideProperties:261 expected:<66536> but was:<65536>
```

while previous:

```
[00:03:28] java.lang.AssertionError: Values should be different. Actual: true
[00:03:28] at uk.gov.nationalarchives.droid.profile.ProfileManagerImplTest.testOverrideProperties(ProfileManagerImplTest.java:243)
```

jcharlet:
Appveyor build was failing on new unit test.

### solution
I added 10ms delay between each test

### issue 

we instantiate a first profile instance, based on current time in ms

```
String profileId = String.valueOf(System.currentTimeMillis());
log.info("Creating profile: " + profileId);
```

then for each assertion we create another profile instance. But on appveyor it's fast enough than time hasn't moved

appveyor logs:

```
2020-07-06T10:37:12,247  INFO [main] ProfileManagerImplTest:235 - start test testOverrideProperties
2020-07-06T10:37:12,247  INFO [main] ProfileManagerImpl:119 - Creating profile: 1594031832247
2020-07-06T10:37:12,247  INFO [main] ProfileContextLocator:112 - creating new ProfileInstance with 
 propertiesOverride 
2020-07-06T10:37:12,247  INFO [main] ProfileManagerImplTest:244 - Test profile.processZip
2020-07-06T10:37:12,247  INFO [main] ProfileManagerImpl:119 - Creating profile: 1594031832247
2020-07-06T10:37:12,247  INFO [main] ProfileContextLocator:115 - ProfileInstance result is already existing
```

11:42
local logs:

```
2020-07-06T11:33:24,034  INFO [main] ProfileManagerImplTest:235 - start test testOverrideProperties
2020-07-06T11:33:24,096  INFO [main] ProfileManagerImpl:119 - Creating profile: 1594031604095
2020-07-06T11:33:24,096  INFO [main] ProfileContextLocator:112 - creating new ProfileInstance with propertiesOverride 
2020-07-06T11:33:24,102  INFO [main] ProfileManagerImplTest:244 - Test profile.processZip
2020-07-06T11:33:24,103  INFO [main] ProfileManagerImpl:119 - Creating profile: 1594031604103
2020-07-06T11:33:24,104  INFO [main] ProfileContextLocator:112 - creating new ProfileInstance with propertiesOverride 
```

jcharlet:
now travis is failing because it cannot install java 9 and java 12 

```
[31;1mThe command "~/bin/install-jdk.sh --target "/home/travis/openjdk9" --workspace "/home/travis/.cache/install-jdk" --feature "9" --license "GPL" --cacerts" failed and exited with 56 during .[0m
```

PR #443, opened 2020-04-06
Add Code Coverage to the build

adamretter:
This adds JaCoCo code coverage to the build.

As an example of use, when running `mvn site` you will now get a `jacoco/index.html` file inside the `target/site` folder of each module (e.g. `droid-core`). This shows comprehensive code coverage of tests.

In addition this adds support for coveralls.io which allows you to easily view the Code Coverage and changes through time in a nice web GUI. See: https://coveralls.io/github/digital-preservation/droid

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 *.war
 *.ear
 
+target/*
 /*/target/
 
 .idea/**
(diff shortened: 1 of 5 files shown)


adamretter:
@jcharlet Some modules are pretty well tested e.g. > 70%. It's the ones that are less than 10% that bring down the average. droid-core has reasonable coverage, and that's probably the most important. So take the figures with a grain-of-salt.

PR #447, opened 2020-04-16
fix: put back LF (unix) line separators in csv exports

### LINKED ISSUE #440: profiles produced now have Windows ends (CR+Newline) rather than just Unix (Newline)
Reported via external testing - is causing issues where DROID is chained with other scripts handling DROID's output

jcharlet:
# Issue 
Fix #440 

# Context 
this is new in 6.5.

This is due to a change in the library used to create CSV format (done by @adamretter to improve processing speed).
The library by default outputs files in a different format according to the operating system (using the operating system's own convention), while the previous one just used Unix format.

```
https://github.com/uniVocity/univocity-parsers/blob/master/src/main/java/com/univocity/parsers/common/Format.java
 *  <li><b>lineSeparator:</b> the 1-2 character sequence that indicates the end of a line. Newline sequences are different across operating systems. Typically:
 *		<ul>
 *			<li>Windows uses carriage return and line feed: <i>\r\n</i></li>
 *			<li>Linux/Unix uses line feed only: <i>\n</i></li>
 *			<li>MacOS uses carriage return only: <i>\r</i></li>
 *		</ul>
 *   	<i>{@link Format#lineSeparator} defaults to the system line separator</i>
 *  </li>
```

# Solution
manually set the CSV format back to Unix.

# Acceptance tests
- start GUI in Windows
- run analysis on random files
- export CSV
- verify that CSV files use LF (Unix) line separating instead of CRLF (Windows)

### DIFF
--- a/droid-export/src/main/java/uk/gov/nationalarchives/droid/export/CsvItemWriter.java
+++ b/droid-export/src/main/java/uk/gov/nationalarchives/droid/export/CsvItemWriter.java
@@ -45,6 +45,7 @@ import org.apache.commons.lang.time.FastDateFormat;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.univocity.parsers.csv.CsvFormat;
 import com.univocity.parsers.csv.CsvWriter;
 import com.univocity.parsers.csv.CsvWriterSettings;
 import com.univocity.parsers.common.TextWritingException;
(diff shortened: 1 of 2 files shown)


jcharlet:
argh sorry @adamretter , you can ignore whitespace changes.
Screenshot from 2020-04-16 15-32-03


will undo that while fixing the checkstyle issue.

jcharlet:
meh.. intellij doesn't see the space change :( 
Screenshot from 2020-04-16 15-35-23

jcharlet:
@adamretter finally understood what the issue was.. and funny enough, it was the way my IDE was formatting the file using.... Unix LF instead of Windows CRLF, which changed the whole files. after a big of fighting with it, I fixed it, checkstyle issues and unit tests on windows. Should be fine now.

adamretter:
Are you using Intellij? If so it normally follows whatever your git repo is configured for I think

jcharlet:
yes but for some reason if there was already files staged, it wouldn't see the change when I update the file format (with intellij or unix2dos shell command).
Can't update your requested changes @adamretter cause I overwrote my commit to keep it clean, but it should be fixed now and ready for review.

Issue #439, opened 2020-04-03
Command line errors in testing


Francesca4242:
I just want to make sure that these errors are standard and nothing unusual. I think they are from corrupted files but you might want to check that?
[log-6.5-command line.txt](https://github.com/digital-preservation/droid/files/4426979/log-6.5-command.line.txt)

jcharlet:
Thanks Francesca for raising this.

I am putting those errors directly in the comment to ease readability.

# errors in text file: to catch and log as WARN
3 types of issues raised in CLI while testing on skeleton suite v96, which could be made quiet.

```
2020-04-03T11:58:42,329 ERROR [core-thread-10] Ole2IdentifierEngine:132 - No reader was obtained for file://na-stor01/homedir$/fmackenzie/Desktop/DROID%20Testing/skeleton-suiteV96/fmt/fmt-111-signature-id-170.nul. This may be due to low memory conditions. Try running with a larger heap size!
```

```
Apr 03, 2020 12:03:39 PM com.github.junrar.Archive setFile
WARNING: exception in archive constructor maybe file is encrypted, corrupt or support not yet implemented
java.lang.NullPointerException
        at com.github.junrar.Archive.readHeaders(Archive.java:291)
        at com.github.junrar.Archive.setFile(Archive.java:169)
        at com.github.junrar.Archive.setVolume(Archive.java:648)
        at com.github.junrar.Archive.<init>(Archive.java:124)
        at com.github.junrar.Archive.<init>(Archive.java:100)
        at uk.gov.nationalarchives.droid.core.interfaces.archive.RarArchiveHandler.handle(RarArchiveHandler.java:103)
        at uk.gov.nationalarchives.droid.submitter.SubmissionGateway.handleArchive(SubmissionGateway.java:269)
        at uk.gov.nationalarchives.droid.submitter.SubmissionGateway.access$700(SubmissionGateway.java:76)
        at uk.gov.nationalarchives.droid.submitter.SubmissionGateway$SubmissionFutureTask.done(SubmissionGateway.java:610)
        at java.base/java.util.concurrent.FutureTask.finishCompletion(Unknown Source)
        at java.base/java.util.concurrent.FutureTask.set(Unknown Source)
        at java.base/java.util.concurrent.FutureTask.run(Unknown Source)
        at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source)
        at java.base/java.util.concurrent.FutureTask.run(Unknown Source)
[...]

PR #448, opened 2020-04-20
nio File based POIFS is optimized for memory, 

sparkhi:
so separated out the 2 error handling by whether we use file or InputStream.

This will make sure that we do not arbitrarily suggest increasing the memory. My analysis of the code indicated that the error message was introduced many years back when we used only `InputStream` this constructor's documentation indicated that it is more memory intensive and there are suggestions to use the constructor which takes in a `File`
We subsequently introduced that constructor (which uses nio) but the memory error message was kept in as it is. This PR separates the two. 

Helps with one part of https://github.com/digital-preservation/droid/issues/439

### DIFF
--- a/droid-command-line/src/main/java/uk/gov/nationalarchives/droid/command/DroidCommandLine.java
+++ b/droid-command-line/src/main/java/uk/gov/nationalarchives/droid/command/DroidCommandLine.java
@@ -65,6 +65,11 @@ public final class DroidCommandLine implements AutoCloseable {
 
     /** Options message. */
     public static final String USAGE = "droid [options]";
+
+    /**
+     * Message about incorrect syntax.
+     */
+    public static final String CLI_SYNTAX_INCORRECT = "Incorrect command line syntax: %s";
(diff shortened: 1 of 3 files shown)


jcharlet:
can't see any breaking change, but I don't understand your description of the PR @sparkhi . You simply moved the instantiation of the input stream in the try block right?

jcharlet:
travis build in error, for some reason, got stuck on jdk 10 and 11, probably because of timeout (ran for 6h). can't restart,

sparkhi:
### Added the command line syntax error handling. 

### Acceptance Tests 
1) POIFS - Mainly through code review. The code is changed such that, the existing error message about memory is used only if we use the stream. If we use a file, the real underlying error is bubbled up as it is.

2) Run CLI  without any options, you should see a message indicating the syntax is incorrect. Try running the CLI with some other syntactical errors (e.g. use -Nr but do not supply a signature file), such errors will also be shown as syntax error message but there should not be any stack trace on the console

PR #487, opened 2020-06-24
No database profiling

jcharlet:
Matt's started work related to https://github.com/digital-preservation/droid/issues/375

PR #489, opened 2020-06-25
No database profiling

### LINKED ISSUE #375: No Profile Mode - issues and questions.
The No Profile mode in the command line project suffers from multiple issues.

- doesn't match extensions the same as profile mode.
- only outputs filename and puid.
- archive file processing support is limited (it usually can't handle nested archives)

**Origin**
The original idea of no profile mode was to support outputting droid metadata without the overhead of a database.  If the results were just going to go to the standard out, or a file, then there was no need to record them in a database as well.

**Implementation**
A decision was made to re-implement all the core identification and archive handling in the command line project for the no profile mode.  This was a short term decision made due to some annoying incompatibilities (filters being one).  At the time there was only zip, gz and tar archive handling.

**Outcomes**
The result of this is that there is a considerable maintenance burden imposed by the no profile mode, as the archive handling must be re-implemented every time a new type is added.  In fact, the implementations provided by no profile mode are much less capable than the ones already available in the main droid projects (they can't handle nested archives for the most part).

Identification also suffers, as the logic doesn't match what is available in the rest of DROID, and results suffer as very little metadata is outputted in this mode (just filename and puid).

**Original design**
In the original design, the idea of no profile mode was that it should use all the same profile machinery, but just plug in a csv output ResultsHandler (whether to standard out or a file), instead of the database results handler.

**Questions**
1.  Would implementing no profile mode to the original design be of interest?  Doing so would massively simplify the command line project, as it would just use the standard droid modules used by the rest of DROID.  
2.  Do people depend on the current output (just filename and puid) of no profile mode?  This could be preserved, although I suspect most people will want the full metadata most of the time.

_Note_: I see there is an issue with no-profile-mode filenames containing commas here: https://github.com/digital-preservation/droid/issues/34  This would be fixed by using a CSV output format for droid metadata.  The idea would be that no profile mode could output data essentially the same as if we had profiled with a database, and then performed an export.  The output could still be given in the current format for backwards compatibility reasons.

jcharlet:
WIP
fix #375
reused branch from @nishihatapalmer on #487, rebased master on this branch

### DIFF
--- a/droid-results/src/main/java/uk/gov/nationalarchives/droid/profile/ProfileContextLocator.java
+++ b/droid-results/src/main/java/uk/gov/nationalarchives/droid/profile/ProfileContextLocator.java
@@ -64,6 +64,7 @@ public class ProfileContextLocator {
     private static final String HIBERNATE_CREATE = "hibernate.hbm2ddl.auto";
     private static final String BLANK_PROFILE = "profile.template";
     private static final String SIG_PROFILE = "profile\\.\\d+\\.template";
+    private static final String OUTPUT_FILE_PATH = "profile.outputFilePath";
     
     private final Logger log = LoggerFactory.getLogger(getClass());
     private DroidGlobalConfig globalConfig;
@@ -125,6 +126,7 @@ public class ProfileContextLocator {
                     .getLong(DroidGlobalProperty.MAX_BYTES_TO_SCAN.getName()));
             profileInstance.setMatchAllExtensions(globalConfig.getProperties()
                     .getBoolean(DroidGlobalProperty.EXTENSION_ALL.getName()));
+            profileInstance.setOutputFilePath(globalConfig.getProperties().getString(OUTPUT_FILE_PATH));
             return profileInstance;
         }
     }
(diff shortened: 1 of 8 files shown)


nishihatapalmer:
I've pushed a few more commits for this to create the right objects using spring.

https://github.com/nishihatapalmer/droid/tree/NoDatabaseProfiling

nishihatapalmer:
In terms of roadmap, we'll need a few things to happen in the right order.

This branch needs a bit more work to add the outputFilePath to the ProfileInstance as a persisted profile property.

We'll then need to use the Profile Properties PR - but we'll need to update it given a conflict with No Database Profiling (they touch some of the same classes, but both come of master).  

At that point, we should have all the machinery to actually do no database profiling from the command line itself.  I guess I can make a local branch which merges both no database profiling AND profile properties before they're merged into master so I can work on those changes.

jcharlet:
Brilliant,
I just checked out your last commit. I'll rename it (because it says fix checkstyle issues, while it's about adding the outputFilePath to the ProfileInstance as a persisted profile property).
I'll work on merging your PR and fix the build.
then I'll merge or rebase master with that PR back onto your branch (I'm working on digital-preservation repo rather than yours to avoid breaking anything on your own repo, in case you have any pending work)

jcharlet:
related PR on properties: https://github.com/digital-preservation/droid/pull/407

Issue #505, opened 2020-08-04
DPP file profiling fails with "Null Byte in Path" error 


sparkhi:
There are a few DPP files which fail when run with Droid 
An example file to reproduce this behaviour is attached. The DPP file is inside the attached zip file. 

[SerifDrawPlusX3-Sample1.zip](https://github.com/digital-preservation/droid/files/5022339/SerifDrawPlusX3-Sample1.zip)

**Implementation Suggestions:**
DPP is handled using `TrueZipArchiveHandler`. This particular file in question has a null byte at the end of each entry name. As the null byte is at the end, it is safe to trim it for profiling purpose. The error is thrown by the `FileNameUtils.getName()` method, it would be easy to trim the name before passing it to the FileNameUtils. 

Make sure corresponding implementation works in CLI as well

Issue #482, opened 2020-06-16
CLI command to give list of available reports returns lots of duplicated lines


jcharlet:
~/Doc/d/d/droid6.5 ❯ ./droid.sh -l                                                                                                           took 20m 38s at 16:08:12
2020-06-16T16:08:19,232  INFO [main] DroidCommandLine:140 - Starting DROID.

Report:	'Total unreadable files'
	Formats:	'Text'	'Web page'	'Pdf'	'DROID Report XML'

Report:	'Total unreadable files'
	Formats:	'Text'	'Web page'	'Pdf'	'DROID Report XML'
Report:	'File count and sizes'
	Formats:	'Text'	'Web page'	'Pdf'	'DROID Report XML'

Report:	'Total unreadable files'
	Formats:	'Text'	'Web page'	'Pdf'	'DROID Report XML'
Report:	'File count and sizes'
	Formats:	'Text'	'Web page'	'Pdf'	'DROID Report XML'
Report:	'File count and sizes by file extension'
	Formats:	'Text'	'Web page'	'Pdf'	'DROID Report XML'

Report:	'Total unreadable files'
	Formats:	'Text'	'Web page'	'Pdf'	'DROID Report XML'
Report:	'File count and sizes'
	Formats:	'Text'	'Web page'	'Pdf'	'DROID Report XML'
Report:	'File count and sizes by file extension'
	Formats:	'Text'	'Web page'	'Pdf'	'DROID Report XML'
Report:	'Total count of files and folders'
	Formats:	'Text'	'Web page'	'Pdf'	'DROID Report XML'

Report:	'Total unreadable files'
	Formats:	'Text'	'Web page'	'Pdf'	'DROID Report XML'
Report:	'File count and sizes'
	Formats:	'Text'	'Web page'	'Pdf'	'DROID Report XML'
Report:	'File count and sizes by file extension'
	Formats:	'Text'	'Web page'	'Pdf'	'DROID Report XML'
Report:	'Total count of files and folders'
	Formats:	'Text'	'Web page'	'Pdf'	'DROID Report XML'
Report:	'File count and sizes by month last modified'
	Formats:	'Text'	'Web page'	'Pdf'	'DROID Report XML'

Report:	'Total unreadable files'
	Formats:	'Text'	'Web page'	'Pdf'	'DROID Report XML'
Report:	'File count and sizes'
	Formats:	'Text'	'Web page'	'Pdf'	'DROID Report XML'
[...]

PR #526, opened 2020-09-17
Bug fixes to remove duplicated lines and trim null character from zipEntry

sparkhi:
2 Changes in here for 2 small bugfixes 
1) #482 - The StringBuilder was not being reinitialised, hence with every report iteration it added more and more details and printed it all
2) #505 - DPP files have a trailing null character in the names, trimming that would help get away from the error.


adamretter:
@sparkhi Any tests?

sparkhi:
> @sparkhi Any tests?

Hey @adamretter I had not written any because the change was too small (initialization and trimming) and the classes did not have any test classes already. I'll try and add one which will need some more code separation in there.

sparkhi:
@adamretter Added test for the report reporting, please take a look. Thanks

PR #566, opened 2021-01-19
Profile id collision 553

sparkhi:
Changes to fix name collision


droid-results/src/main/java/uk/gov/nationalarchives/droid/profile/ProfileManagerImpl.java
```diff
@@ -141,6 +141,23 @@ public ProfileInstance createProfile(Map<SignatureType, SignatureFileInfo> sigFi
...
+    private String getUniqueProfileId() {
+        String profileId = String.valueOf(System.currentTimeMillis());
+        Path profileHomeDir = config.getProfilesDir().resolve(profileId);
+
+        //to avoid folder collision, check if the folder already exists, if it does, just wait and try again
+        while (FileUtil.exists(profileHomeDir)) {
+            try {
+                Thread.sleep(1);
+            } catch (InterruptedException ex) {
+                //ignored - unlikely to happen
```
adamretter, line 153: This loop has the potential to go on for ever. I would suggest two things:

1. When you can the InterruptedException, you restore the interrupt flag on the current thread (otherwise you risk issues elsewhere)
2. You only retry a finite number of times.

sparkhi: We are getting the currentTimeMillis, which would give us a new time and hopefully a folder with that name does not exist. Not entirely sure in what scenario it would go on forever? 
I could change it to run a finite number of times but then the question is, what if we do not get a valid profileId? Should it throw an exception?

sparkhi: Done the suggested changes, @adamretter let me know your comments

PR #620, opened 2021-06-03
Move CI from Travis to GitHub Actions

adamretter:
Travis CI is closing down its .org service and severely limiting its credits for its free .com service. As such it is a better idea to move over to GitHub Actions these days.


adamretter:
Any change of some review/merge action?

sparkhi (approved):
slight difference in the JDK versions we used on travis and here, but not major to stop the PR.

Issue #580, opened 2021-02-15
BUG: command line filter parser cannot handle strings with more than one space.


nishihatapalmer:
If you try to filter a profile with a filename with more than one space in it, we get an exception.  It can handle a name with a single space, using:

```droid -p "/home/user/myprofile.droid" -E "/home/user/myprofile.csv" -f "file_name = 'a name'"```

Whereas this fails:

```droid -p "/home/user/myprofile.droid" -E "/home/user/myprofile.csv" -f "file_name = 'a name with several spaces'"```

This is because of the regex in SimpleSqlFilterParser that splits up the parameters.  After one space, it terminates the parameter, and puts the rest of the filename into another parameter.

I will fix this in my work on the no profile mode replacement.

nishihatapalmer:
This is now fixed in the no profile mode replacement code (PR yet to be submitted).

sparkhi:
Tested in the current release candidate, it works. 👍🏼

Issue #578, opened 2021-02-13
Set default signature from command line can only set binary signatures


nishihatapalmer:
The code can only update the binary signatures at present using the set command:

```droid -s 97```

will set binary signature version 97, but:

```droid -s 20200121```

will fail to find the container signature with that version and set it.

I've made a small change in the command line work I'm doing to allow both binary and container signature defaults to be set.

Issue #572, opened 2021-02-03
SUGGESTION: Quiet mode by default in command line


nishihatapalmer:
Quiet mode in the command line turns off logging to the console.   I suggest this should be the default.

There is almost no circumstance where you want to see the internal workings on DROID on the command line, mixed up with the actual results the command line is outputting, since they both output to the same place.

Most command line utilities do the opposite - they have a verbose mode if you want to see what's going on for debug purposes - otherwise, they just output the results of running a command to the console.  This is for good reasons, as if you want to further process the results, having log entries mixed up in the very data you're trying to work with makes this much harder.  Or to just see what the results are as a human.

The current situation is rather like if the GUI mixed up the log entries with the tree view on the main screen!

nishihatapalmer:
There is only thing other than logging that quiet mode affects: in the No Profile mode, it also outputs some information about signatures and the run, unless quiet mode is specified.

No profile mode is being replaced by the work I'm doing on outputting to CSV directly from a profile (without writing to the database).  So there is a question on whether we want to keep supporting the existing No Profile Mode behaviour.  If we don't care about writing this header out, then quiet mode ONLY affects outputting logging to the console and nothing else.

nishihatapalmer:
I have made a change in my no profile code.  Logging exists to the console where the command is not itself outputting to the console (e.g. when writing to a report file, export or a profile database).  When the command's purpose is to output something to the console, logging to console is disabled.

sparkhi:
Tested with the profile export command (which does not have output to console) vs signature updated (which has an output to console) and they seem to work fine.

Issue #486, opened 2020-06-23
inconsistency in archives content identifications


sparkhi:
https://trello.com/c/1A2pined/75-inconsistency-in-archives-content-identifications #486

The classes for management of each archive format exist in the GUI and the code is copied into the CLI (for no-profile mode only). The aim is to keep one copy in common module and consume it from both places. 

[UPDATE Jeremie] add link in description

nishihatapalmer:
I've now got the command line using the profile machinery for all identification work, but without writing to the database and instead outputting directly to CSV.  Once tested and accepted, we can remove all the existing command line code for archive file processing and file type identification.  The GUI and the command line will share the same code for this.

rhubner:
Fixed in #743

PR #631, opened 2021-06-12
Simplify and enhance command line code

nishihatapalmer:
This PR massively simplifies the command line, by removing all the duplicated handling of archives, filters and signature matching.  The command line project now only parses commands and then uses standard machinery outside of it.  This will result in a much lower maintenance burden and avoids discrepancies between how command line and GUI operate.

There have been a lot of slow careful changes to get to this point.  While the number of commits may seem large, please also observe the final state of the code.  It is much smaller, simpler and comprehensible.  The code has been extensively tested.  Through code removal and the addition of new tests, the command line project now has 88% line coverage in unit tests.  I have also run a set of manual test cases for all the command line options.  Test result spreadsheet is attached below

[6.6 test results.xlsx](https://github.com/digital-preservation/droid/files/6642624/6.6.test.results.xlsx)

Along the way I have also made some additions to the command line capabilities.  In order to preserve the output formatting of the old "no profile mode", it is now possible to configure what columns are exported, and whether they are quoted or not.  Filtering has become more powerful, allowing list of extensions, and the ability to filter out files before they are identified on the basis of file metadata.  All identification commands can output results directly to CSV format on the console or to a file.   The help file for command line control has been completely rewritten.  

All changes are fully backwards compatible in terms of command structure; no existing commands are affected and will continue to operate as before.   The new capabilities are enabled with new command options (or for console output, by not specifying a profile or file to output to).  There is a small change to log output to console and no profile mode (see backward compatibility comment below).

This PR addresses the issues:
https://github.com/digital-preservation/droid/issues/580
https://github.com/digital-preservation/droid/issues/578
https://github.com/digital-preservation/droid/issues/572
https://github.com/digital-preservation/droid/issues/486
https://github.com/digital-preservation/droid/issues/471
https://github.com/digital-preservation/droid/issues/470
https://github.com/digital-preservation/droid/issues/375


nishihatapalmer:
Looks like a have a language feature that's too high for version 8 and 9.... will investigate.

nishihatapalmer:
The two failing checks appear to be something to do with the signature service checks.  Nothing here has changed - I seem to remember this signature service test depends on network connectivity and may fail with timeouts occasionally.

nishihatapalmer:
Somewhat confused - now there are apparently conflicts when there weren't before - and the code had been merged with upstream master.   Many of those conflicts listed above are in files that are now removed from the command line project (one of the main points of this PR!)

adamretter:
@nishihatapalmer This sounds like a great idea. Just one question, is it 100% backwards compatible with the options that the user has currently. i.e. if a user has scripted DROID, are the same command line options available with the same syntax as before?

adamretter:
@nishihatapalmer Regards any conflicts, I can assist you in rebasing this if you would like?

nishihatapalmer:
In terms of backwards compatibility, existing commands will continue to function precisely as before, with two exceptions to the output:

**Quiet mode.** 
This change only affects informational commands whose entire purpose is to write information to the console (e.g. help, version, list reports).  They would typically write out an "INFO Starting DROID" message as the first line.  Note there is no guarantee that this will be the only thing logged to console - that just depends on whether the rest of the DROID code chooses to log what it's doing.  So any script that relies on say, stripping out the first line could fail at any time.

To include arbitrary log messages mixed with the actual command output makes the command almost impossible to rely on for automated processing. 

If complete backwards compatibility is desired, I'd be tempted to create a property or environment variable that could flip the behaviour between "log arbitrary messages to console with my command output" and "don't log to console if I'm trying to write info to the console!" modes!  DROID could ship with the old behaviour, but have a switch for those of us who like sane command line applications!

**No profile mode.**  
This used to have a sort of "metadata header" it wrote out to the console including what the signature file versions were and whether extensions were being filtered.  I have not put this header back in for the current code.  If it is deemed to be required, I could add it back again.  

Happy to discuss what the compatibility requirements are and make any changes as required.

nishihatapalmer:
I've merged master with my branch, resolved conflicts and fixed license header files (for files I've introduced).

nishihatapalmer:
The remaining build failures seem to be a problem of the build environments being unable to download an owasp file.

nishihatapalmer:
### New capabilities
An overview of the main new capabilities in this PR.  

* Specify columns to export with the `-co` option, e.g. `-co NAME PUID`
* Configure CSV to only quote if there is a comma in a field, use the `-qc` option
* Write a profile straight to a CSV file, not writing to a database, using the `-o` output option, e.g. `-o "/home/user/results.csv"
* Write a profile straight to the console, by not specifying an output file or a profile to write to.
* You can filter profiles written to CSV file or console just as for an export.
* Filter a list of extensions, e.g. `-F "file_ext any doc docx xls xlsx"`.  Previous version could only filter on single extensions.
* Filter files before they are identified on file metadata (name, extension, last modified, size), using the `-ff` and `-FF` filter options.  If there are files you don't want to process, it is faster to filter them before identification takes place, particularly if they are embedded in an archive file, as it avoids the need for a lot of file I/O.

The majority of these capabilities were added specifically to replicate existing functionality in "no profile mode", which always behaved slightly differently (e.g. only outputted a couple of columns) but had some useful features (e.g. extension filtering).  So essentially this is transferring the features required to support no profile mode features into the main code base.

nishihatapalmer:
The Appveyor build failure I'm struggling to understand.  It says a particular test `DroidCommandLineTest.testRunAndSaveProfileNoArgs()` is failing as the arguments are different.  But as far as I can see, the arguments are identical as they should be, and the test just passes on my machine.

Could this be something to do with line endings on windows or unix?  The test assumes \n for new lines.

adamretter:
> Could this be something to do with line endings on windows or unix? The test assumes \n for new lines.

Sounds likely. Should probably use `System.getProperty("line.separator")` instead of just assuming `\n`.

nishihatapalmer:
> > Could this be something to do with line endings on windows or unix? The test assumes \n for new lines.
> 
> Sounds likely. Should probably use `System.getProperty("line.separator")` instead of just assuming `\n`.

Well, that seems to have fixed it.  I just limited the test to the contents of the first line, which is fine.

droid-command-line/src/main/java/uk/gov/nationalarchives/droid/command/action/ProfileRunCommand.java
```diff
@@ -107,19 +114,52 @@ public void onProgress(Integer progress) {
         
     }
 
+    /**
+     * @return The default binary and container signatures, but will override with different ones if provided.
+     * @throws SignatureFileException if there's a problem obtaining the signature files.
+     */
+    private Map<SignatureType, SignatureFileInfo> getSignatureFiles() throws SignatureFileException {
+        Map<SignatureType, SignatureFileInfo> sigs = signatureManager.getDefaultSignatures();
+        if (binarySignaturesFileName != null) {
```
sparkhi, line 123: binarySignatureFile could be empty, if it is set as empty from outside of this class, then `Paths.get` might throw an exception

nishihatapalmer: True.  Or if it simply has an incorrect file name.  `Paths.get()` would throw an `InvalidPathException`, which is a subtype of RuntimeException, and propagate up to the `DroidCommandLine.processExecution()` method, which would catch it and output an error to the console.

Of course, the error may not be particularly helpful to the user.  A custom exception if we can't get a Path could provide better user feedback.

nishihatapalmer: The point of the null test isn't actually to avoid Path errors, btw.  If a specific signature file is not set, then we use the profile defaults which are already obtained.  If *something* is set (not null), then we try to use that instead.     I'm working on a tiny change that gives better user feedback if a supplied filename is invalid.

sparkhi: That's cool, thanks 👍🏼

nishihatapalmer: Turns out it is impossible from the command line to set a blank signature file.  And we get an error if you try to use a non-blank but invalid signature file.  So I would argue we might not need to do anything here.

droid-results/src/main/java/uk/gov/nationalarchives/droid/results/handlers/WriterResultHandlerDao.java
```diff
@@ -0,0 +1,177 @@
...
+
+    @Override
+    public synchronized void init() {
+        try {
+            super.init();
+        } finally {
+            if (writer == null) { // If no writer is set, default to console output.
+                writer = new PrintWriter(System.out);
+            }
+            itemWriter.open(writer);
```
sparkhi, line 114: Do we not need to close this itemWriter

sparkhi (approved):
The changes look good to me

droid-results/src/main/java/uk/gov/nationalarchives/droid/results/handlers/WriterResultHandlerDao.java
nishihatapalmer: Would be better to close it I think, good catch.  There's no close() method in any of the results handling interfaces, so would need to add those and wire it all up.  I'll have a look at doing that.

nishihatapalmer: Turns out adding close() is harder than it looks.  Since processing multi-threaded, if you close a profile as soon as you've finished walking the profile resources, there can still be identification requests in flight on other threads.  This can cause the underlying connections (file or database) to close before all identification is finished, throwing various errors and failing to record the last profile results.

nishihatapalmer: Would need to have some kind of thread safe signal that everything was finished in order to allow safe close.

nishihatapalmer: I may have a look at this - but I don't think this is a major issue.  What would be the worst effect of not closing the CSV writer and underlying file access?  On the command line, execution ceases as soon as the profile is complete, so there won't exactly be any resource leak.  This writer can't be invoked from the GUI (which writes to database).

nishihatapalmer: Some investigation says that the primary effect of not explicitly closing something is that there may be unflushed data which may not get written.  The current `CsvItemWriter` code calls `flush()` after each file it writes out so there is no risk currently we will not write all data out.

nishihatapalmer: It does make me wonder though what happens when the command line closes down.  If it closes down as soon as all profile resources are finished, what does that do to threads that are in flight?  Could we lose the last few identifications?  Doesn't seem to have been a problem for writing to the database when profiling on the command line, but might bear a bit more investigation.

adamretter:
@sparkhi @nishihatapalmer are you both still keen on getting this in? If so I could try and rebase it?

nishihatapalmer:
Yes, I'm keen to get this in.  It will simplify all future maintenance of
DROID and make it consistent, as the command line will use the same code as
everything else.

adamretter:
@sparkhi Any comments?

rhubner:
@adamretter @nishihatapalmer I fix conflicts. You can see it there: https://github.com/digital-preservation/droid/tree/NoDatabaseProfilingSimplifyFilter

Let me know what do you think.

nishihatapalmer:
Looks like all these merges were just fixing upgrades to various components
that have happened since the PR was published, and formatting (checkstyle?)
type changes to formatting and javadoc.

Was there anything in particular that worried you in the merge?

rhubner:
> Was there anything in particular that worried you in the merge?
@nishihatapalmer Only the size of PR. But from communication which I can see it should be ok.

rhubner:
@adamretter Yes. Everythinkg what is in `digital-preservation/droid:NoDatabaseProfilingSimplifyFilter` Is prepared for merge.

The best will be if @nishihatapalmer can pull all changes and update his branch. This should solve all conflicts on this PR.

rhubner:
Merged as #743

PR #654, opened 2021-09-28
Ignored a couple of tests that refer to non-existent pronom endpoint

sparkhi:


rhubner:
Closing thi PR. All fixed in #740

PR #675, opened 2021-12-22
Updated to use Open HTML TO PDF and removed itextpdf

kparmar1:
Updated to use Open HTML TO PDF and removed itextpdf due to vulnerability (CVE-2021-43113)

### DIFF
--- a/droid-help/src/main/resources/Web pages/Third-party components.html
+++ b/droid-help/src/main/resources/Web pages/Third-party components.html
@@ -376,6 +376,16 @@
       </tr>
       <tr>
         <td align="center" width="50%">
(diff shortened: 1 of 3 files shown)


DavidUnderdown:
Thanks Kal

All the devs are now off for Christmas I think, so someone will look at this properly on their return.

sparkhi:
Hi Kal,
I've created a branch and merged your changes into that branch in droid. I'll go through the changes locally and do some tests as the code change is small but fundamental to report generation and then merge it into master. 
Thanks a lot for your contribution

kparmar1:
Np, I'm glad to have been able to contribute.

PR #682, opened 2022-01-11
Fixed broken link

thearchdefect:
I noticed that the link to download the latest binary file from The National Archives website was broken (404), so this PR should fix it.


adamretter:
@thearchdefect Thanks very much for your contribution :-)

PR #724, opened 2022-03-08
Dependencies upgrade

rhubner:


adamretter:
I think @DavidUnderdown and @Dclipsham may have previously had concerns about how backwards compatible the situation is when upgrading the version of Derby. Do you know if there are any issues?

rhubner:
@adamretter Good question. I think it should be forward compatible, but don't know if backward. I will double check.

sparkhi (approved):

rhubner:
@adamretter 
It looks like it's not backward compatible. 
`.droid6/profiles/1646924277862/db has an incompatible format with the current version of the software.  The database was created by or upgraded by version 10.14.`

I will remove DerbyDB update

PR #758, opened 2022-04-12
Add simple API for another TNA team

rhubner:
* This is internat API and should not be used outside TNA. There is no guarantee it will stay here.


droid-results/src/main/java/uk/gov/nationalarchives/droid/internal/api/ContainerApi.java
```diff
@@ -0,0 +1,126 @@
...
+
+        try {
+            zip.init();
+        } catch (SignatureFileException e) {
+            throw new RuntimeException("Unable to init zip identifier", e);
+        }
+        return zip;
+    }
+
+    private Ole2IdentifierEngine ole2IdentifierEngine() {
```
sparkhi, line 106: Is this one line method necessary?

sparkhi (changes requested):
Looks good, except one minor question inline 
and, it would be good to have some javadoc on the API class / methods that would benefit the consumer of the API

droid-results/src/main/java/uk/gov/nationalarchives/droid/internal/api/ContainerApi.java
rhubner: Maybe not. I just what to have it same. I will remove.

PR #804, opened 2022-08-11
dependency upgrades

tballison:
Dependency upgrades with required updates for checkstyle and maven dependency plugin.


sparkhi:
Thanks for contributing. At this moment, I'll close this PR without merging. Quite a few of dependency upgrades are automatically handled by dependabot but a few changes in this PR are good to look out for in future

PR #815, opened 2022-09-07
updated tna logo on splash screen

sparkhi:
logo updated on splash screen to use the current TNA logo instead of the old 'A'


OliverHannan (approved):
Thanks

Issue #805, opened 2022-08-18
sigtool behaviour with inverted ranges


steve-daly:
If you look at PRONOM fmt/142 the raw signature is `52494646{4}57415645666D7420[!10]{3}[!FEFF]{16-*}64617461`

This is decomposed within the PRONOM binary signature file as:

   ```
<ByteSequence Endianness="Big-endian" Reference="BOFoffset">
        <SubSequence Position="1" SubSeqMaxOffset="0" SubSeqMinOffset="0">
          <Sequence>57415645666D7420</Sequence>
          <LeftFragment MaxOffset="4" MinOffset="4" Position="1">52494646</LeftFragment>
          <RightFragment MaxOffset="0" MinOffset="0" Position="1">[!10]</RightFragment>
          <RightFragment MaxOffset="3" MinOffset="3" Position="2">[!FEFF]</RightFragment>
        </SubSequence>
        <SubSequence Position="2" SubSeqMinOffset="16">
          <Sequence>64617461</Sequence>
        </SubSequence>
      </ByteSequence>
```

If I use sigtool to generate the XML instead, I get this...

```
<ByteSequence Reference="BOFoffset">
  <SubSequence Position="1" SubSeqMaxOffset="0" SubSeqMinOffset="0">
    <Sequence>57415645666D7420</Sequence>
    <LeftFragment MaxOffset="4" MinOffset="4" Position="1">52494646</LeftFragment>
    <RightFragment MaxOffset="0" MinOffset="0" Position="1">10</RightFragment>
    <RightFragment MaxOffset="3" MinOffset="3" Position="2">[00:FD]</RightFragment>
  </SubSequence>
  <SubSequence Position="2" SubSeqMaxOffset="16" SubSeqMinOffset="16">
    <Sequence>64617461</Sequence>
  </SubSequence>
</ByteSequence>
```

Note that for the first RightFragment, the value has become '10' rather than [!10] thereby inverting the logic. 

This can be reproduced more simply by running commands like:

sigtool [!10]
sigtool [!10:12]

which give the results:
10
[10:12]

...without the necessary exclamation mark, and thereby inverting the logic. 

Although I mention sigtool here, I believe this is calling core DROID code. 

Note that fmt/142 includes the troublesome and ambiguous [!FEFF] string, but this issue isn't related to that.  

It was suggested that @nishihatapalmer might be interested in this behaviour.

nishihatapalmer:
Thanks - I'll have a look at it.

[...]

PR #816, opened 2022-09-16
Fix serialization - inverted bytes and byte ranges

nishihatapalmer:
This PR fixes the inverted bytes and byte ranges reported in #805.

It also fixes some minor formatting issues with spacing of some elements, putting hex digits in uppercase to be consistent, and adds tests for the standard syntax that can be serialized.


droid-core/src/test/java/uk/gov/nationalarchives/droid/core/signature/compiler/ByteSequenceSerializerTest.java
```diff
@@ -103,12 +465,11 @@ private void testRenderSignatures(String fileName, SignatureType type) throws Ex
...
                     compileExpression(sequence, type);
-                } catch (CompileException e) {
-                    failureMessages += e.getMessage();
-                    failed = true;
-                } catch (ParseException e) {
-                    failureMessages += e.getMessage();
+                } catch (CompileException | ParseException e) {
+                    failureMessages += (sequence + " : " + e.getMessage());
                     failed = true;
+                    compileExpression(sequence, type);
```
sparkhi, line 471: For the test, once we have gathered that it failed, should we just move on to next expression, not sure about the purpose of `compileExpression` in the catch block

droid-core/src/test/java/uk/gov/nationalarchives/droid/core/signature/compiler/ByteSequenceSerializerTest.java
```diff
@@ -103,12 +465,11 @@ private void testRenderSignatures(String fileName, SignatureType type) throws Ex
             if (sequence.trim().length() > 0) { //ignore blank lines in the test input files.
```
sparkhi, line 465: defence against possible empty lines 👍🏼 
Should we also defend against possible blank space characters (i.e. would it be better to trim the sequence before passing it into `compileExpression`?)

sparkhi (commented):
A couple of minor questions in the test class, otherwise looks good.

droid-core/src/test/java/uk/gov/nationalarchives/droid/core/signature/compiler/ByteSequenceSerializerTest.java
nishihatapalmer: The purpose of this test is to read all the signatures in the file (which is all of them that existed when the test was written), to ensure that we get no compilation exceptions when attempting to render them as different types of signature.

We gather all the failures together into a single message before failing the test, so we can see patterns in failures and not just play whack-a-mole as each individual one fails in turn.

nishihatapalmer: Sorry - I just realised what you were asking.

That should be taken out.  It was left in when I was debugging a failed expression - so I could put a breakpoint in after the failure.

nishihatapalmer: I've submitted a fix for this, removing the compileExpression() call.

droid-core/src/test/java/uk/gov/nationalarchives/droid/core/signature/compiler/ByteSequenceSerializerTest.java
nishihatapalmer: I put this in because I occasionally left a blank line or spaces in the test files.  This just ensures we're always testing an actual expression - a line with something in it.

droid-core/src/test/java/uk/gov/nationalarchives/droid/core/signature/compiler/ByteSequenceSerializerTest.java
sparkhi: Thanks very much. 👍🏼

droid-core/src/test/java/uk/gov/nationalarchives/droid/core/signature/compiler/ByteSequenceSerializerTest.java
sparkhi: Thanks 👍🏼

sparkhi (approved):

Issue #821, opened 2022-09-30
sigtool behaviour with ?? wildcard


steve-daly:
If I use _sigtool_ to create a Pronom style signature from a byte sequence thus:

`sigtool  -b -p "04{1}[01:0C][01:1F]{28}([41:5A]|[61:7A]){10}(43|44|46|4C|4E)" `

The results splits correctly into the XML form...

```xml<ByteSequence Reference="BOFoffset">
  <SubSequence Position="1" SubSeqMaxOffset="0" SubSeqMinOffset="0">
    <Sequence>04</Sequence>
    <RightFragment MaxOffset="1" MinOffset="1" Position="1">[01:0C][01:1F]</RightFragment>
    <RightFragment MaxOffset="28" MinOffset="28" Position="2">[41:5A]</RightFragment>
    <RightFragment MaxOffset="28" MinOffset="28" Position="2">[61:7A]</RightFragment>
    <RightFragment MaxOffset="10" MinOffset="10" Position="3">43</RightFragment>
    <RightFragment MaxOffset="10" MinOffset="10" Position="3">44</RightFragment>
    <RightFragment MaxOffset="10" MinOffset="10" Position="3">46</RightFragment>
    <RightFragment MaxOffset="10" MinOffset="10" Position="3">4C</RightFragment>
    <RightFragment MaxOffset="10" MinOffset="10" Position="3">4E</RightFragment>
  </SubSequence>
</ByteSequence>
```

but if I swap the {1} for ?? which should be syntactically identical, the processing doesn't split a fragment at the ?? like it does with {1} and the question marks pass through into the XML output, which DROID then hangs when processing. 

`sigtool  -b -p "04??[01:0C][01:1F]{28}([41:5A]|[61:7A]){10}(43|44|46|4C|4E)" `

```xml<ByteSequence Reference="BOFoffset">
  <SubSequence Position="1" SubSeqMaxOffset="0" SubSeqMinOffset="0">
    <Sequence>04</Sequence>
    <RightFragment MaxOffset="0" MinOffset="0" Position="1">??[01:0C][01:1F]</RightFragment>
    <RightFragment MaxOffset="28" MinOffset="28" Position="2">[41:5A]</RightFragment>
    <RightFragment MaxOffset="28" MinOffset="28" Position="2">[61:7A]</RightFragment>
    <RightFragment MaxOffset="10" MinOffset="10" Position="3">43</RightFragment>
[...]

PR #822, opened 2022-10-09
FIXES: Any ?? matcher should not be part of sequences.

nishihatapalmer:
?? symbols were being incorrectly included as part of the XML signature text.  They should instead be modelled as min/max offsets between what precedes them and what follows them.  This fix removes the ANY (??) matcher as a valid part of an anchoring sequence, as that was incorrect.

The fix is to ensure that the ANY type is not processed as part of the anchor text, and cannot be included in fragment text either.  Finally, sigtool correctly takes account of the anchor type (BOF/EOF) when serializing signatures.

sigtool will now correctly process `-b -p "04??[01:0C][01:1F]{28}([41:5A]|[61:7A]){10}(43|44|46|4C|4E)"`, with the ?? turned into min/max offset of 1.


nishihatapalmer:
Fixes #821

sparkhi (approved):

PR #870, opened 2023-01-19
Changed the Windows batch file to use the embedded jre

paulwellsagilej:


droid-binary/bin/sigtool.bat
```diff
@@ -1,2 +1,37 @@
...
+@REM NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+@REM SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+@REM
+
 @echo off
-java -jar droid-tools-${project.version}.jar %*
+
+SET DROID_HOME=%~dp0
+
+start "" "%DROID_HOME%jre\bin\java.exe" java -jar droid-tools-6.6.0-rc3-SNAPSHOT.jar %*
```
sparkhi, line 37: Please change to use `${project.version}`  instead of hardcoded version name

sparkhi (approved):

droid-binary/assembly.xml
```diff
@@ -98,23 +98,11 @@
...
     </file>
     <file>
-      <source>bin/Running DROID.txt</source>
+      <source>bin/sigtool.bat</source>
       <outputDirectory>/</outputDirectory>
       <filtered>true</filtered>
       <fileMode>644</fileMode>
     </file>
-    <file>
-      <source>bin/droid.sh</source>
```
steve-daly, line 107: This looks like it removes the sigtool and droid.sh scripts from the multi-platform package. Is that intentional? These files are needed for non-Windows users.

Issue #861, opened 2023-01-06
Change to the CSV export


CaptainBrad:
Currently, if you ask for 'one row per file', any additional identified formats are added as extra columns - but they are not given a column header.  Can this be included, e.g. PUID2 etc.  The missing header makes the csv difficult to use in Excel and also for scripting where you have to work around the missing headers (e.g. Python pandas read_csv).  I'm using 6.5.2.

ross-spencer:
@CaptainBrad this looks like it's connected to: https://github.com/digital-preservation/droid/issues/118 there's some additional discussion there.

CaptainBrad:
Thanks for the heads up @ross-spencer , we will look into 118.

paulwellsagilej:
I understand the problem now and I have investigated and started making changes to add the missing header values. The number of columns will vary from one CSV file to another, but each file will be valid within itself. The number of values in each row will match the number of values in the header row.

paulwellsagilej:
Existing tests pass now. Working on new tests to check new behaviour.

PR #873, opened 2023-01-26
Ticket 861 csv

paulwellsagilej:


steve-daly:
Hi @paulwellsagilej this is title "Ticket 861 csv" but seems to include other commits such as the Windows Batch File changes.  Was that intentional?

steve-daly:
Can I also check about the CSV column heading implementation. In Issue #861 the spec suggested that repeated column heading should have numeric suffixes (e.g. PUID2 as in the example given).  Excel needs unique column headings for some cases (e.g. Pivot Tables, or index lookups) so I think we'd want to avoid having duplicated column headings.

paulwellsagilej:
@steve-daly According to my local comparison between branches I have changed only two files, so I'm investigating why GitHub thinks nearly 1400 files have changed. Probably a fight over line endings as suggested by @sparkhi. Thanks for the nudge about keeping headers unique - I have just pushed a change for that.

sparkhi:
This PR should not be merged yet. There is an issue around bean instantiation using spring, so the changes in this PR would succeed the build but not launch droid.

droid-results/src/main/java/uk/gov/nationalarchives/droid/profile/CsvItemWriter.java
```diff
@@ -29,354 +31,365 @@
...
-
-    private static String toFileName(String name) {
-        return FilenameUtils.getName(name);
-    }
-    
-    /**
-     * No config is needed by this class, but it's retained temporarily for backwards compatibility purposes.
-     * @param config the config to set
-     */
-    public void setConfig(DroidGlobalConfig config) {
```
sparkhi, line 268: Removal of this setter causes problem in spring config

sparkhi (commented):
There are a few problems that need to be addressed before this PR can be merged 
1) Removal of the setter method for config, makes it fail at runtime as the spring config tries to call this setter method
2) Folders do not have format count and consequently there is a NPE thrown on a folder as we compare for column count
3) When we run the export again, the header for "HASH" is removed but the data is retained which is a critical problem.

sparkhi:
The work is done through other branch

PR #882, opened 2023-02-06
Fixed the problem with the default output format for no-profile

paulwellsagilej:
Listing the file path rather than the unqualified name


steve-daly:
What's the behaviour with archive containers (e.g. zip, tar, gzip files).  Does this now give the same output as 6.5.2 for these cases too?

PR #891, opened 2023-02-23
Correct filename for bzip archive

sparkhi:
This PR has multiple changes 
1) Correct filename scheme for bZip archive
2) Correction for the default processing of -A, -W, -At and -Wt flags in case of no-profile mode
3) Some simpler coding improvements flagged up by warnings and checkstyles


paulwellsagilej (approved):
All looks ok to me

PR #901, opened 2023-03-06
issue-900 fix for hash column missing from the export

sparkhi:
Fix for issue 900.
As we were copying reference of the HEADER, then subsequently modifying the HEADER to include a prefix based on algorithm, on the second run, it was complaining about a header mismatch between data and headers. Making a deep copy of the HEADERS solves the column name mismatch


droid-results/src/main/java/uk/gov/nationalarchives/droid/profile/CsvItemWriter.java
```diff
@@ -107,6 +107,7 @@ public class CsvItemWriter implements ItemWriter<ProfileResourceNode> {
     private static final int MIME_TYPE_ARRAY_INDEX          = 15;
     private static final int FORMAT_NAME_ARRAY_INDEX        = 16;
     private static final int FORMAT_VERSION_ARRAY_INDEX     = 17;
+    private static final String BLANK_SPACE_DELIMITER = " ";
```
steve-daly, line 110: Do you think you could tweak the formatting of this constant to match the rest of the block of definitions, i.e. align the equals sign with the others.

sparkhi: done and pushed the change

steve-daly (approved):

PR #930, opened 2023-04-04
Moved the header writing closer to writing the data so we can decide …

sparkhi:
…on headers when writing per file


sparkhi:
abandoning this

PR #933, opened 2023-04-05
For GUI and profile, moved the header writing closer to data writing …

sparkhi:
…so that we can generate additional headers when exporting per file
(Merging to master rather than patch branch)

The main changes are

* Moved writing of headers closer to when the data is about to be written - hence the unit test for checking that the header writing happens in "open" has been removed
* Look through the data to see if we need to create additional headers "per format", if so, create those headers
* Do the additional headers ONLY if we are writing the records "one file per row"
* If a single row is written, do not attempt to write headers again, this is necessary for defence in the no-profile mode where the CSVWriter is invoked once per resource.

### DIFF
--- a/droid-results/src/main/java/uk/gov/nationalarchives/droid/profile/CsvItemWriter.java
+++ b/droid-results/src/main/java/uk/gov/nationalarchives/droid/profile/CsvItemWriter.java
@@ -52,12 +52,15 @@ import java.net.URI;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.Date;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 /**
  * @author rflitcroft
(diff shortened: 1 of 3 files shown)


sparkhi:
The main changes are 
1) Moved writing of headers closer to when the data is about to be written - hence the unit test for checking that the header writing happens in "open" has been removed 
2) Look through the data to see if we need to create additional headers "per format", if so, create those headers 
3) Do the additional headers ONLY if we are writing the records "one file per row" 
4) If a single row is written, do not attempt to write headers again, this is necessary for defence in the no-profile mode where the CSVWriter is invoked once per resource.

droid-results/src/main/java/uk/gov/nationalarchives/droid/profile/CsvItemWriter.java
```diff
@@ -220,7 +256,43 @@ public void open(final Writer writer) {
         if (headers == null) {
             headers = Arrays.copyOf(HEADERS, HEADERS.length) ;
         }
-        String[] headersToWrite = getHeadersToWrite(headers);
+    }
+
+    private void writeHeadersForOneLinePerFormatExport(List<? extends ProfileResourceNode> nodes) {
```
kurtismash, line 261: Is this function name correct, should it not be `writeHeadersForOneRowPerFileExport`? My understanding is that if a file has >1 formats this function will add up to 4 extra columns for each additional format but each file will still only have 1 row.

droid-results/src/main/java/uk/gov/nationalarchives/droid/profile/CsvItemWriter.java
```diff
@@ -220,7 +256,43 @@ public void open(final Writer writer) {
...
+
+        if (options == ExportOptions.ONE_ROW_PER_FORMAT) {
```
kurtismash, line 263: Is there a better way to write this check that would extend to cover additional options in future? I can see that `if (options != ExportOptions.ONE_ROW_PER_FILE)` wouldn't work at the moment as it's possible for no option to be set. Could the switch on line 176 be refactored to set the option `ONE_ROW_PER_FILE` if the default case occurs?

sparkhi: Actually its a good point, I had specifically moved the option setting line in "ExportTask" so the negative condition would be better

droid-results/src/main/java/uk/gov/nationalarchives/droid/profile/CsvItemWriter.java
sparkhi: That's correct. I'll rename

droid-results/src/main/java/uk/gov/nationalarchives/droid/profile/CsvItemWriter.java
```diff
@@ -220,7 +256,43 @@ public void open(final Writer writer) {
...
+                    }
+                    if (headersToWrite.contains(HEADER_NAME_MIME_TYPE)) {
+                        headersToWrite.add(HEADER_NAME_MIME_TYPE + newColumnSuffix);
+                    }
+                    if (headersToWrite.contains(HEADER_NAME_FORMAT_NAME)) {
+                        headersToWrite.add(HEADER_NAME_FORMAT_NAME + newColumnSuffix);
+                    }
+                    if (headersToWrite.contains(HEADER_NAME_FORMAT_VERSION)) {
+                        headersToWrite.add(HEADER_NAME_FORMAT_VERSION + newColumnSuffix);
+                    }
```
kurtismash, line 292: Is it worth putting another loop here to iterate over `PER_FORMAT_HEADERS` to reduce the duplication? Readability would take a hit.

droid-results/src/main/java/uk/gov/nationalarchives/droid/profile/CsvItemWriter.java
```diff
@@ -162,6 +190,9 @@ public void write(List<? extends ProfileResourceNode> nodes) {
     }
     
     private void writeOneRowPerFile(List<? extends ProfileResourceNode> nodes) {
+        if (csvWriter.getRecordCount() == 0) {
+            writeHeadersForOneLinePerFileExport(nodes);
```
kurtismash, line 194: Sorry for being pedantic, could this be `row` instead of `line` as everything else in this file refers to lines as rows?

kurtismash (approved):

PR #940, opened 2023-04-24
Add file extension mismatch to the ApiResult in internal API

sparkhi:
Added file extension mismatch to the internal API.
* implemented the publishing of file extension mismatch to the internal API
* added a few tests 
* renamed the tests to reflect what they are trying to test 
.
.
... and also updated the page https://github.com/digital-preservation/droid/wiki/How-to-use-DROID-internal-API so please take a look there as well.


droid-api/src/test/java/uk/gov/nationalarchives/droid/internal/api/DroidAPITest.java
```diff
@@ -109,23 +113,38 @@ public void testFileExtensions() throws IOException, SignatureParseException {
...
+    public void should_report_correct_version_for_the_binary_and_container_signature() {
         assertThat(api.getContainerSignatureVersion(), is("20221102"));
         assertThat(api.getDroidVersion(), is(ResourceBundle.getBundle("options").getString("version_no")));
         assertThat(api.getBinarySignatureVersion(), is("109"));
     }
 
     @Test
-    public void testEmptyFileWithNoExtension() throws IOException, SignatureParseException {
-        DroidAPI api = DroidAPITestUtils.createApi();
+    public void should_report_correct_size_of_zero_for_an_empty_file() throws IOException {
```
MancunianSam, line 141: This is just a suggestion but is it worth changing this test name slightly. `size_of_zero` sounds like you're referring to the file size. Maybe something like `should_produce_zero_results_for_an_empty_file` ? Then it matches the test name below as well.

droid-api/src/test/java/uk/gov/nationalarchives/droid/internal/api/DroidAPITest.java
```diff
@@ -31,29 +31,38 @@
  */
 package uk.gov.nationalarchives.droid.internal.api;
 
+import org.junit.Before;
 import org.junit.Test;
+import org.junit.rules.Timeout;
```
MancunianSam, line 36: This import is unused and the `java.util.Arrays` one underneath.

MancunianSam (commented):
Looks good, thanks. Couple of small things.

droid-api/src/test/java/uk/gov/nationalarchives/droid/internal/api/DroidAPITest.java
sparkhi: I was trying to figure out the reason behind the test that runs a loop and at some point I thought, maybe that test was written for performance in mind hence thought of applying a timeout but decided against it, the import remained.

sparkhi: Done

droid-api/src/test/java/uk/gov/nationalarchives/droid/internal/api/DroidAPITest.java
sparkhi: Done

PR #945, opened 2023-05-12
Move zip library from truzip to truevfs

sparkhi:
Truezip is replaced with TrueVFS
1) Changes in POM for dependency
2) Changes in interfaces and results for the identification purpose 
3) Implementation for SeekableByteChannel to make use of underlying reader
4) Change in DiskProfileAction for reading / writing the profile files.


droid-core-interfaces/src/main/java/uk/gov/nationalarchives/droid/core/interfaces/archive/ByteseekWindowWrapper.java
```diff
@@ -31,99 +31,99 @@
...
+        return reader.length();
+    }
+
+    @Override
+    public SeekableByteChannel truncate(long size) throws IOException {
+        throw new IOException(NOT_IMPLEMENTED);
+    }
+
+    @Override
+    public boolean isOpen() {
```
kurtismash, line 125: Is this method used anywhere as I can't see it in the original `TrueZipReader`? Or are we including this for parity with `SevenZipReader`?

If we are keeping it should it have some tests?

kurtismash:
Looks good, just the 1 comment around that unused method.

kurtismash (approved):

PR #954, opened 2023-06-08
Made the arguments to -At and -Wt as optional

sparkhi:
The actual changes related to the ticket is 2 folds 
1) In `CommandLineParam`, the `ARCHIVE_TYPES` and `WEB_ARCHIVE_TYPES` have been changed to make their arguments optional
2)  In `CommandFactoryImpl` if there are no arguments passed to -At or -Wt, the property override is defensive against null

There are tests added for command line without the arguments.

All other changes are code cleanup identified as warnings.

### DIFF
--- a/droid-command-line/src/main/java/uk/gov/nationalarchives/droid/command/action/CommandLineParam.java
+++ b/droid-command-line/src/main/java/uk/gov/nationalarchives/droid/command/action/CommandLineParam.java
@@ -94,7 +94,7 @@ public enum CommandLineParam {
      */
     PROFILE_PROPERTY("Pr", "profile-property", true, -1, I18N.PROFILE_PROPERTY_HELP, "profile property") {
         @Override
-        public DroidCommand getCommand(CommandFactory commandFactory, CommandLine cli) throws CommandLineException {
+        public DroidCommand getCommand(CommandFactory commandFactory, CommandLine cli) {
             return null;
         }
     },
@@ -104,7 +104,7 @@ public enum CommandLineParam {
      */
     PROPERTY_FILE("Pf", "property-file", true, 1, I18N.PROPERTY_FILE_HELP, "property file") {
         @Override
-        public DroidCommand getCommand(CommandFactory commandFactory, CommandLine cli) throws CommandLineException {
+        public DroidCommand getCommand(CommandFactory commandFactory, CommandLine cli) {
             return null;
         }
     },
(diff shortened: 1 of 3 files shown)


droid-command-line/src/main/java/uk/gov/nationalarchives/droid/command/action/CommandLineParam.java
```diff
@@ -315,7 +311,7 @@ public DroidCommand getCommand(CommandFactory commandFactory, CommandLine cli) {
         }
     },
     /** Open webarchive types. */
-    WEB_ARCHIVE_TYPES("Wt", "open-webarchive-types", true, 2, I18N.WEB_ARCHIVE_TYPES_HELP, "web archive types") {
+    WEB_ARCHIVE_TYPES("Wt", "open-webarchive-types", false, 2, I18N.WEB_ARCHIVE_TYPES_HELP, "web archive types") {
```
MancunianSam, line 314: Feel free to ignore this but is it worth putting a test in to say that the `argsRequired` is now false. There is a `CommandLineParamTest` class but there aren't many tests in it. Is it worth adding a couple of tests for these two?

MancunianSam (commented):
Looks good thanks. Just a couple of questions.

droid-command-line/src/main/java/uk/gov/nationalarchives/droid/command/action/CommandLineParam.java
sparkhi: Well spotted. We use the GNU parser which throws an exception if required argument is true. As the test code is using the parser directly (i.e. not passing it to the implementation for parsing) there is not much point in testing the parameter change. The desired result is covered by the other tests.  
The problem that the test code is using the parser directly is beyond the scope of this ticket

MancunianSam: Ok, makes sense, thanks.

droid-command-line/src/test/java/uk/gov/nationalarchives/droid/command/action/CommandFactoryTest.java
```diff
@@ -1324,6 +1317,72 @@ public void no_profile_mode_should_set_specific_archive_types_to_expand_when_hyp
...
+        ProfileRunCommand e1 = (ProfileRunCommand) factory.getNoProfileCommand(cli);
+
+        assertNotNull(e1);
+        assertFalse(e1.getRecursive());
+
+        assertEquals("stdout", e1.getDestination()); // output file not specified, so defaults to stdout.
+        List<String> propertyNames = getPropertyNames(e1.getProperties());
+        assertEquals(13, propertyNames.size());
+
+        // -Only the 3 archives passed to -At should expand
```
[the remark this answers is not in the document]
sparkhi: Perils of copy paste, changed now

MancunianSam (approved):

PR #955, opened 2023-06-13
Ability to use SHA512 for checksum implementation done

sparkhi:
SHA512 algorithm made available. The profile db template is updated to allow for longer length of the hash column


kurtismash (approved):
Looks good

PR #984, opened 2023-06-21
Prep for new release - signature files and skeleton suite

sparkhi:
PR for preparing for 6.7.0 release. The PR looks big but the changes are straightforward and only related to tasks we do for a version release:
 
1) Update droid to use the latest signature file - and update any tests accordingly
2) Update the skeleton suite (corpus of small made up file of each format) for the tests to correspond with the latest embedded signature file
3) Updates to changelog and documentation to include the items for new version


kurtismash (approved):
Looks good. I wonder if it's worth investigating the use of symlinks to reduce the duplication of signature files in future releases.

PR #1038, opened 2024-01-04
Added a suppression for derby database

sparkhi:
As per https://nvd.nist.gov/vuln/detail/CVE-2022-46337
We have a dependency that we use with high vulnerability, however the vulnerability is associated with using LDAP for login. We do not use LDAP for our derby database (the db that stores the results), hence suppressing the check is easier than rebuilding the jar ourselves or move to a newer version of java.

While at it, also removed all the spring vulnerabilities that had anyways expired in 2022 and we had moved to a newer version of spring.


steve-daly (approved):
Happy with this.  No issue supressing the LDAP-related vulnerability as we're not using this authentication method.

PR #1084, opened 2024-04-03
fixes memory leak on large files

rsehr:
This PR fixes a memory leak with large (50GB+) files

### DIFF
--- a/droid-tools/src/main/java/uk/gov/nationalarchives/droid/tools/SigUtils.java
+++ b/droid-tools/src/main/java/uk/gov/nationalarchives/droid/tools/SigUtils.java
@@ -749,9 +749,10 @@ public final class SigUtils {
         long lastmodified = Files.getLastModifiedTime(file).toMillis();
         RequestMetaData metaData = new RequestMetaData(size, lastmodified, file.toString());
         RequestIdentifier identifier = new RequestIdentifier(file.toUri());
-        IdentificationRequest fileRequest = new FileSystemIdentificationRequest(metaData, identifier);
-        fileRequest.open(file);
-        return new IdentificationRequestByteReaderAdapter(fileRequest);
+        try (IdentificationRequest fileRequest = new FileSystemIdentificationRequest(metaData, identifier)) {
+            fileRequest.open(file);
+            return new IdentificationRequestByteReaderAdapter(fileRequest);
+        }
     }
 
     /**


rsehr:
I had problems with any large mxf video file. They were between 50GB and 250GB in size, so they are too big to simply upload somewhere.

With an old droid version (6.1.3) I had no problems, but after the upgrade to the latest version these files could no longer be processed because of OutOfMemoryException. By closing the FileSystemIdentificationRequest object and the underlying WindowReader object, the files can be processed again.

sparkhi:
I assume, it launches Droid as a command line interface? 
In that case, there is an option `-Pr "profile.maxBytesToScan=<some-max-bytes-value>"`  
e.g. `-Pr "profile.maxBytesToScan=65536"` this would restrict Droid to read only 65536 bytes from the beginning or end of file for identification purpose.

steve-daly:
@rsehr Do you see this issue both with and without checksum generation enabled (remembering you need to make a new profile each time after changing this setting).  Does DROID take a long time to characterise the first large MXF file, even without Checksum generation enabled?

I've just tried this (on a Mac) using a range of really large MXF files but don't see any significant memory usage or leakage, even when making checksums.   

Could you upload the details from the 'About' box. 

We'd just like to reproduce the issue as it's not something we've seen before and we do work with large video files.

rsehr:
I checked the nanite code, they use the droid java api from what I can see. I prepared a test class to reproduce it. It uses the same api calls as the nanite library:

```
public class TestMain {

    private static String filename = "/path/to/large/file.mxf";
    private static Path containerSignatureFile =
            Paths.get("/path/to/container-signature-20230307.xml");
    private static Path droidSignatureFile = Paths.get("/path/to/DROID_SignatureFile_V111.xml");

    public static void main(String[] args) throws Exception {
        DroidAPIExtended api = DroidAPIExtended.getInstance(
                droidSignatureFile,
                containerSignatureFile,
                Long.MAX_VALUE);
        api.setAllowMatchByFileExtension(false);
//        api.setMaxBytesToScan(65536);
        InputStream input = Files.newInputStream(Paths.get(filename));

        List<ApiResultExtended> ids = api.submit(input, filename);
        ApiResultExtended r = ids.get(0);
        String mimeTypeString = r.getMimeType();
        System.out.println(mimeTypeString);
    }
}
```

It doesn't matter, if the line setMaxBytesToScan is commented in or not. The result is in both cases an OutOfMemoryError.

sparkhi:
Thanks for the sample code @rsehr 
One thing that is slightly baffling me is, the original PR you suggested is in `droid-tools` and the InternalApi call, which nanite has modified does not look like it touches the `droid-tools` module. 

I'll try and get hold of a large MXF file here to see if I can reproduce it in a similar setup that you have. I might not be able to get to it imminently, I'll try at some point. Thanks

sparkhi:
@rsehr 
Looking at nanite code, it uses a different `submit` method (`List<ApiResultExtended> ids = api.submit(input, filename);`) which has a memory leak. I tried running a 12GB MXF file in loop, multiple times and set the max heap to a low value. I managed to see the failure after about 10 iterations.
When I ran the exact same thing with Droid (without nanite), using the only public `submit` method available on the internal api (`api.submit(Paths.get(<path to the file>));`) I did not experience the memory leak.

I would like to close this pull request.

For interest, the memory leak in nanite has a potential small fix -> https://github.com/openpreserve/nanite/pull/46

anjackson:
Hi @rsehr, and thanks for looking at this @sparkhi. Unfortunately, this is my fault, in that this is really a documentation bug. The docs do not clearly state the assumption that _if you open an InputStream you are responsible for closing it_.  This is because, in the context I used Nanite, it was very important that the passed InputStream was _not_ closed, because the calling code needed to re-use it. I'll try and improve the Nanite documentation to make this clear, as the docs don't mention it, and the example code doesn't close the InputStreams either!  My apologies for the time wasted on this.

sparkhi (commented):
As per the discussions above, this PR is not needed

anjackson:
I just want to note that, as far as I can tell, the change given in the PR does not change any of the code associated with the Nanite InputStream-based identification code. Maybe there is a resource leak here, in the `getByteReaderForFile(Path file)` code that identifies files passed in by Path. I don't think it has anything to do with the InputStream version.

Note that, if possible, users should always use the File-based version if they can. The InputStream version is intended for cases where there isn't a filesystem, and may well blow up on large files in some cases. Again, I'll try to make the documentation clearer.

PR #1086, opened 2024-04-10
Migrated the signature file download client from obsolete version to current

sparkhi:
The container signature files are updated by using a simple "REST api" 
We used httpClient from apache common which has medium vulnerability and also it has been superseeded by another incarnation

This PR updates to use the latest version.


techncl (approved):
Looks good!

PR #1113, opened 2024-06-11
Improved the validation error messages on export template parsing

sparkhi:
Merging the error message improvements into the release branch
The main change is, 
Earlier the order of checks was 
1) Is it a profile results column?
2) Is it a constant string column?
3) Anything else - treat it as Data modifier column 

Now we have changed the checking order between 2 and 3, this means "anything else" is treated as a constant string column and automatically, the associated error messages indicate syntax error accordingly


MancunianSam (approved):
:+1:

PR #1148, opened 2025-01-14
DR2-2038 Upate Derby database

MancunianSam:
This is now the latest version of derby.

This version needs Java 21. I've updated the bundled version to use 21.

The name of the driver class has changed and has been updated in all of
the places it's set.

I wanted to get this working on 21 anyway. DROID itself runs happily on
21 but the tests weren't passing because of how old Mockito was so I've
upgraded that.

I've removed versions 7 and 11 and 17 from the CI workflow.

I've updated the ci workflow to use an NVD API key which is set in the
repository secrets to stop the ci run taking an hour periodically.

I've removed the supressions for the old version of the derby DB

The last thing was an odd test failure in the tests for ProfileFileChooser
It was saying that multiSelectionEnabled was true when run on Java 21.
If you run DROID on Java 21 then multi selection is enabled. I'm not
sure why but setting it to false in the setup method brings back the old
behaviour.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,14 +13,13 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        jdk: [8, 11, 17]
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v2
       - name: Set up JDK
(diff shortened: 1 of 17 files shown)


techncl (approved):
👍 . Wow, having the CI not taking an hour would is great!

sparkhi (approved):
Looks good to me

PR #1151, opened 2025-01-29
Migrate from javax.xml to jakarta

MancunianSam:
This needs doing as it's holding us back upgrading some libraries to the
latest version.


droid-container/pom.xml
```diff
@@ -102,8 +102,8 @@
             <artifactId>xercesImpl</artifactId>
         </dependency>
         <dependency> <!--JAXB-API-->
```
sparkhi, line 104: The comment is now wrong. Actually, the comment was a bit pointless as it just says the "artifactId", it can simply be removed.

droid-core-interfaces/pom.xml
```diff
@@ -88,8 +88,8 @@
         </dependency>
 
         <dependency> <!--JAXB-API-->
```
sparkhi, line 90: similar thing about comment

droid-swing-ui/pom.xml
```diff
@@ -116,8 +116,8 @@
     </build>
     <dependencies>
         <dependency> <!--JAXB-API-->
-            <groupId>javax.xml.bind</groupId>
-            <artifactId>jaxb-api</artifactId>
+            <groupId>jakarta.xml.bind</groupId>
```
sparkhi, line 119: similar thing about comment

sparkhi (commented):
just added a couple of comments about removing comments

droid-container/pom.xml
MancunianSam: Good spot, I've removed this one and the other two.

sparkhi (approved):

PR #1165, opened 2025-02-03
DR2 2076 Do not increment loop counter.

### LINKED ISSUE #1140: TIFF file(s) causing DROID to hang
The attached TIFF file causes DROID to hang - basically a scan never completes. This isn't the only one - others from the same digitized batch have the same issue. (I've attached within a zip but scanning from within or without the zip has the same hang issue)

I can't see anything obviously broken within the file, but I'm getting no clue from DROID logs as to why it's problematic. It might well be some interesting TIFF subset format, but even if it is, that would typically still give me a result and I could look at improving PRONOM.
[RGE0021-B23F26-0001-008.zip](https://github.com/user-attachments/files/17932226/RGE0021-B23F26-0001-008.zip)

Environment details:
DROID 6.8, Windows 10, Java 11 OpenJDK as supplied in the DROID binary

David

MancunianSam:
DROID is getting stuck in an infinite loop trying to match the right fragments for [this signature](https://github.com/nationalarchives/pronom-signatures/blob/develop/signatures/fmt/134.json#L132)

The steps are something like:

1. It finds matches for 7 of the 8 right fragments.
2. It adds these to a matched list
3. It doesn't find a match for the 8th fragment
4. It then searches each of the fragments in the matched list to either side of the previous matches.
5. Fragments 1-7 match and so stay in the matched list
6. The last matched fragment position is set to 7
7. The loop checking each fragment increments the fragment position to 8
8. It doesn't find a match for the 8th fragment
9. Repeat steps 3 - 9

The problem I think is step 6. It shouldn't be changing the loop counter.

The issue is that I'm not 100% sure if by doing this, we'll be skipping
some fragment checks which will cause a different format to not be
matched at some point in the future.
I don't think so though because the main loop still iterates through
every fragment.


sparkhi (approved):
Its quite a core change, I think we will need quite a bit of testing with variety of files

PR #1216, opened 2025-03-05
Fix the JSON output.

MancunianSam:
As it was, it was only outputting the last file if there was more than one in the profile.

This was because we were creating the `OutputJson` object for each file we were writing which created the json array node as new every time so we only ever got the last item.

I've changed this so it's only created once. 

This then caused the problem that every intermediate json object was being written out. I've fixed this by using the `JsonGenerator` which streams each object to the writer. 

This led to another problem in that a json array needs `[` and `]` wrapping it. It's simple to add one at the beginning. Adding one at the end for the gui export is fine because we know when the export is complete but it's more tricky with the CLI because all of the nodes in the profile run in different threads.

I added a method in the `ProfileRunCommand` which writes `]` to the end of either the PrintWriter or FileWriter if the output is set to JSON. It's not elegant but it does work.


droid-command-line/src/main/java/uk/gov/nationalarchives/droid/command/action/ProfileRunCommand.java
```diff
@@ -100,18 +102,30 @@ public void onProgress(Integer progress) {
...
+
+    public Writer getWriter(String outputFilePath) throws IOException {
+        if (outputFilePath != null && !outputFilePath.trim().isEmpty()) {
+            File outputFile = new File(outputFilePath);
+            return new FileWriter(outputFile, true);
+        }
+        return new PrintWriter(System.out);
+    }
+
+    public void terminateJsonArray() throws IOException {
```
sparkhi, line 122: just a suggestion, to write a comment for this method's purpose, so others will know

droid-help/src/main/resources/Web pages/Command line control.html
```diff
@@ -230,6 +241,20 @@ <h3>
 <p>
     <i>Note:</i> if you specify "stdout" as the filename for the -o option, it will also write to the console.  But it's easier to simply not specify the -o option.
 </p>
+<p>
+    To output JSON results to a file rather than CSV, you can use the <b>--json</b> option:
```
techncl, line 245: Should this be "To output results to a JSON rather than CSV..." or are you actually saying it produces JSON results that can be output to a CSV/JSON?

droid-help/src/main/resources/Web pages/Command line control.html
```diff
@@ -1355,7 +1400,16 @@ <h3> <a id="OptionsTable" name="OptionsTable">Command options</a></h3>
...
             <p>
-            droid -s 20200121
+            droid
+        </td>
+    </tr>
+    <tr>
+        <td>-json</td>
+        <td>--json-output</td>
+        <td>{none}</td>
+        <td>Sets the output to json. This will work for any command which outputs to console or to a file</td>
```
techncl, line 1410: "JSON" is lower cased here but elsewhere it is upper cased

droid-help/src/main/resources/Web pages/Command line control.html
MancunianSam: No, you're right, this doesn't make sense. I think what I meant was
`To output results as JSON instead of CSV, you can use the <b>--json</b> option:`

droid-command-line/src/main/java/uk/gov/nationalarchives/droid/command/action/ProfileRunCommand.java
```diff
@@ -100,18 +102,35 @@ public void onProgress(Integer progress) {
...
-            throw new CommandExecutionException(e);
-        } catch (SignatureFileException e) {
-            throw new CommandExecutionException(e);
         }
-        
+
+    }
+
+    public Writer getWriter(String outputFilePath) throws IOException {
+        if (outputFilePath != null && !outputFilePath.trim().isEmpty() && !"stdout".equals(outputFilePath)) {
```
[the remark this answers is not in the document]
MancunianSam: 1. Yeah, why not. I think the argument parser does it but there's no harm doing it here.
2. I was going for the old-school avoid an NPE by doing `.equals` on the String literal but we're checking for null first so there's no need. I've swapped it round. 
3. No, I think it's fine. If I do then I'll have to call it `trimmedOutputFilePath` or similar and I don't want to :smile:

droid-help/src/main/resources/Web pages/Command line control.html
MancunianSam: I've changed it to upper case.

techncl (approved):
👍

PR #1305, opened 2025-04-30
UI and CLI tests

MancunianSam:
The UI tests use AssertJ Swing to test a running version of the UI.

The tests are executed inside a docker container which allows them to
run on GitHub runners where there is no UI set up.

The CLI tests use bats and run the CLI through a terminal rather than
calling the classes like the unit tests do.

These are meant to be tests made before a dev release to make sure
nothing is seriously broken. They are *not* meant to be completely
comprehensive, nor are they a replacement for manual testing before a
full release.


cli-tests/test/test.bats
```diff
@@ -0,0 +1,245 @@
+#!/usr/bin/env bats
```
techncl, line 1: Maybe the file name could be a little more descriptive; "cli-test" maybe?

cli-tests/test/test.bats
```diff
@@ -0,0 +1,245 @@
...
+
+setup_file() {
```
techncl, line 3: Is there a more descriptive name for this method like "setup_binary_file", "retrieve_script" or "unzip_script"?

techncl (commented):
Looks good! Just a few minor comments, as well as:

1. For the toolbar, do you think it's worth making the options (in DroidMainFrame) Enums and then using them in the tests instead of strings?
2. What about adding tests for:
    1. Start button
    2. A testing to see if columns sort when you click them
    3. What about testing the filter button
    5. Check for signature update
    6. Install signature file
    7. Throttle
    8. Help
    9. I would've added "Expand" and "Expand next three levels" but I can't tell what these do

droid-swing-ui/src/test/java/uk/gov/nationalarchives/droid/gui/ui/SwingUiTestUtils.java
```diff
@@ -0,0 +1,293 @@
...
+    static final GenericTypeMatcher<JLabel> saveProfileMatcher = new GenericTypeMatcher<>(JLabel.class) {
+        @Override
+        protected boolean isMatching(JLabel component) {
+            return "Saving profile...".equals(component.getText());
+        }
+    };
+
+    void waitForProgressLabelInvisible(JLabelFixture labelFixture) {
+
+        Condition condition = new Condition("Wait for invisible initialising progress bar label") {
```
techncl, line 92: Should this be `"Wait for invisible 'initialising' progress bar label"`?

droid-swing-ui/src/test/java/uk/gov/nationalarchives/droid/gui/ui/SwingUiTestUtils.java
```diff
@@ -0,0 +1,293 @@
...
+    };
+
+    static final GenericTypeMatcher<JLabel> saveProfileMatcher = new GenericTypeMatcher<>(JLabel.class) {
+        @Override
+        protected boolean isMatching(JLabel component) {
+            return "Saving profile...".equals(component.getText());
+        }
+    };
+
+    void waitForProgressLabelInvisible(JLabelFixture labelFixture) {
```
techncl, line 90: Maybe `waitForInvisibleProgressLabel`?

MancunianSam:
> For the toolbar, do you think it's worth making the options (in DroidMainFrame) Enums and then using them in the tests instead of strings?

That's a fairly large change and it doesn't massively improve things. It also couples the tests even more to the DroidMainFrame class which I'd like to avoid. As I've said before, I try not to touch the working internals too much with DROID in case it has a knock on effect else where. 

> What about adding tests for:

I put in the PR comment that this isn't meant to be completely comprehensive and some of these are too fiddly or flaky to work reliably.

Start button - There are a few tests already for this. 
A testing to see if columns sort when you click them - Fiddly - hard to check the ordering in the code. 
What about testing the filter button - Fiddly - lots of extra buttons to click.
Check for signature update - I tried this, very flaky. Reliant on the connection to the DROID SOAP API which can be slow/not working at all.
Install signature file - As above. 
Throttle - This affects the timing of the run which is very hard to test.
Help - I'm going to massively change this at some point soon so not now. 
I would've added "Expand" and "Expand next three levels" but I can't tell what these do. - I'm not sure either, sounds fiddly though.

droid-swing-ui/src/test/java/uk/gov/nationalarchives/droid/gui/ui/SwingUiTest.java
```diff
@@ -0,0 +1,564 @@
...
+                .filter(Component::isShowing)
+                .forEach(panel -> utils.waitForProgressLabelInvisible(frame.panel(panel.getName()).label(progressBarMatcher)));
+        Assert.assertEquals(utils.getProfilePanels().size(), existingPanelCount + 1);
+        robot().waitForIdle();
+    }
+
+    @Test
+    public void should_add_files_with_add_files_button() {
+        frame.button("Add File").click();
+        utils.selectFolder("/home/test");
```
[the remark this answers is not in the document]
MancunianSam: I've added "/home/test" to a global variable.

ui-tests/ui-test-startup.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+# Wait until DISPLAY is accessible
+mkdir -p /root/.droid6/profile_templates/schema\ 6.7.0/
+export DISPLAY=:1
+
+echo "Waiting for X to be available..."
```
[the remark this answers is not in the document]
MancunianSam: 1. It might not have to wait 30 seconds though so this isn't quite true. I suppose you could say "up to 30 seconds" but it's fairly clear from the next line. 
2. https://en.wikipedia.org/wiki/X_Window_System

droid-swing-ui/src/main/java/uk/gov/nationalarchives/droid/gui/action/NewProfileAction.java
```diff
@@ -147,7 +147,7 @@ protected void done() {
     
     @Override
     protected Void doInBackground() {
-        
+
```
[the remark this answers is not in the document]
MancunianSam: No, I've removed it.

cli-tests/test/test.bats
MancunianSam: It's part of the framework https://bats-core.readthedocs.io/en/stable/tutorial.html#avoiding-costly-repeated-setups It has to be that.

techncl (approved):
👍

Issue #1274, opened 2025-04-15
Provide "clean" scripts in <userhome>/.droid6 folder


ross-spencer:
I don't think there's an open issue for this, but I think it's a fairly well known issue for signature developers.

Sometimes replacing or adding signature files "confuses" droid when it loads a new profile. I haven't identified the root error yet or if there is something in the logging, I'll need to do some more testing. (It can be a heisenbug on my system). 

If a mechanism can be added to droid itself, that would be great. Otherwise, in the `.droid6` folder, a delete script would be helpful. It can be called by signature developers before reloading droid.

My example for `nix:

```sh
#!/usr/bin/bash

rm -rf logs
rm -rf profile_templates
rm -rf export_templates
rm -rf report_definitions
rm -rf filter_definitions
rm -rf tmp
rm -rf profiles
rm -rf log4j2.properties
```

NB. avoids use of `sudo`.

DROID can be shutdown. `clean.sh` can be run. And then DROID can be restarted. 

Alternative options might be to pinpoint why DROID sometimes wrestles loading new profiles with new signature files and then implement the function in the UI.

While it is easy for individuals to setup a similar or different preferred workflow to achieve the same, providing some tooling for this type of thing represents some shared knowledge that makes it easier for newcomers. 

That being said, tips appreciated if others don't really have this problem and have a better workflow.

ross-spencer:
Connected to: https://github.com/digital-preservation/droid/issues/13#issuecomment-556289691

steve-daly:
Where people are experiencing unexpected behaviour with DROID I normally recommend they delete (or rename) the whole .droid6 folder, but maybe that's not ideal if you have useful test profiles you are working on.  

I can't remember if we mentioned this much, but there was a long-standing bug in DROID meaning that although it intended to delete active profiles on exit, this wasn't working and it left behind profiles which over time slowed down DROID.  We fixed this in 6.8.0 but obviously profiles from before this date wouldn't be touched, and if DROID crashes it doesn't clean up after itself. 

I'd ideally like to track down the root cause of the issue.  If you ever find this reproducible we could investigate, and maybe add some more debugging to see what's going on.  

[...]

PR #1324, opened 2025-05-08
Detect signature file changes.

MancunianSam:
Relates to https://github.com/digital-preservation/droid/issues/1274

When DROID installs a signature file, it creates a profile template. This profile template is a derby database will all of the file formats stored in it.
Once this template exists, DROID will re-use it for any new profiles. This causes the problem where if you change the signature file, DROID won't pick up the changes. You have to delete the ~/.droid6 folder to force DROID to rebuild the profile template database.

Now, when a signature file is loaded, a property is put into droid.properties with the last modified date of the sig file. When DROID starts or when a new profile is created, it checks the current last modified against the one in the properties file. If the sig file has been updated, it returns a status of NO_TEMPLATE and rebuilds the template database.

This now means that if you make a change to the signature file, this will be picked up with any new profile, without having to restart DROID or clean the ~/.droid6 folder.

### DIFF
--- a/droid-results/src/main/java/uk/gov/nationalarchives/droid/profile/ProfileContextLocator.java
+++ b/droid-results/src/main/java/uk/gov/nationalarchives/droid/profile/ProfileContextLocator.java
@@ -33,6 +33,8 @@ package uk.gov.nationalarchives.droid.profile;
 
 import java.io.IOException;
 import java.net.URI;
+import java.net.URLEncoder;
+import java.nio.charset.Charset;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.HashMap;
@@ -41,6 +43,7 @@ import java.util.Properties;
 
 import org.apache.commons.configuration2.CombinedConfiguration;
 import org.apache.commons.configuration2.PropertiesConfiguration;
+import org.apache.commons.configuration2.ex.ConfigurationException;
 import org.apache.commons.configuration2.tree.OverrideCombiner;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -53,6 +56,8 @@ import uk.gov.nationalarchives.droid.export.interfaces.ExportOutputOptions;
 import uk.gov.nationalarchives.droid.results.handlers.JDBCBatchResultHandlerDao;
 import uk.gov.nationalarchives.droid.results.handlers.ProgressObserver;
 import uk.gov.nationalarchives.droid.util.FileUtil;
+
+import static uk.gov.nationalarchives.droid.core.interfaces.config.DroidGlobalProperty.BINARY_SIGNATURE_LAST_UPDATED;
 
 
 /**
@@ -220,6 +225,7 @@ public class ProfileContextLocator {
         if (newDatabase) {
             final Path profileTemplate = getProfileTemplateFile(profile);
             status = getTemplateStatus(profileTemplate);
+            status = checkModificationOfBinarySignature(profile.getSignatureFileName(), status);
             status = setupDatabaseTemplate(status, profileTemplate, databasePath);
         }
         /*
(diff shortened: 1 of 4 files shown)


droid-results/src/test/java/uk/gov/nationalarchives/droid/profile/ProfileContextLocatorTest.java
```diff
@@ -0,0 +1,190 @@
...
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.*;
+
+public class ProfileContextLocatorTest {
+
+    @Test
+    public void testOpenProfileInstanceManagerCreatesNewTemplateIfNewProfile() throws IOException, SignatureFileException, ConfigurationException {
+        String profileId = UUID.randomUUID().toString();
+        Path profileHome = Files.createDirectory(Files.createTempDirectory("profiles").resolve(profileId));
```
sparkhi, line 60: I think, the created directory will be left there instead of getting cleaned up. unless you use `deleteOnExit` on it. 
For junit tests, I prefer to use the TemporaryFolder Rule.
```
    @Rule
    public TemporaryFolder temporaryFolder = new TemporaryFolder();

File: droid-results/src/test/java/uk/gov/nationalarchives/droid/profile/ProfileContextLocatorTest.java [hunk 2]
```diff
@@ -0,0 +1,190 @@
...
+
+        verify(profileInstanceLocator, times(2)).freezeDatabase(eq(profileId));
+        verify(profileInstanceLocator, times(2)).thawDatabase(eq(profileId));
+        verify(globalConfig).update(eq(Map.of("profile.binarySignatureLastUpdated.testsigfile.xml", Files.getLastModifiedTime(signatureFilePath).toMillis())));
+
+        List<String> templateFiles = Files.list(templatesDirectory)
+                .map(Path::getFileName)
+                .map(Path::toString)
+                .toList();
+        Assert.assertArrayEquals(templateFiles.toArray(), new String[]{"profile.1.template", "profile.template"});
```
sparkhi, line 75: This seems to fail for me locally due to order being different. I thought `assertArrayEquals` is independent of order but looks like it is not. 🤔

droid-results/src/main/java/uk/gov/nationalarchives/droid/profile/ProfileContextLocator.java
```diff
@@ -322,6 +328,30 @@ private void generateNewDatabaseAndTemplates(final ProfileInstance profile,
...
+                    long lastModifiedFromConfig = globalConfig.getProperties().getLong(binarySignatureLastUpdatedName);
+                    if (lastModifiedTime > lastModifiedFromConfig) {
+                        globalConfig.update(Map.of(binarySignatureLastUpdatedName, lastModifiedTime));
+                        return TemplateStatus.NO_TEMPLATE;
+                    }
+                }
+            }
+            return status;
+        } catch (ConfigurationException | IOException e) {
+            throw new RuntimeException("Error checking the last modified status of the binary signature file", e);
```
sparkhi, line 352: I know this will be rare situation but, if we can't detect the last modified time, should we stop and throw exception? Or just log this situation as warning and carry on like how we have it currently?

sparkhi (changes requested):
Couple of comments inline

droid-results/src/test/java/uk/gov/nationalarchives/droid/profile/ProfileContextLocatorTest.java
MancunianSam: I've added in a sorted in the stream calls which should force it to be in the order that's there.

droid-results/src/main/java/uk/gov/nationalarchives/droid/profile/ProfileContextLocator.java
MancunianSam: I switched back and forth as to whether this should be a warning or an error and I think you're right, it's a bit risky throwing an error here considering most users won't ever modify the signature file. I've changed it to a `log.warn` call

sparkhi (approved):

PR #1325, opened 2025-05-08
Create 'clean' option for DROID.

MancunianSam:
This will clean the folders out of the ~/.droid6 directory.

This is often what is recommended when DROID misbehaves so having a
command to do this will be useful.

It will delete the following directories and files

container_sigs/
signature_files/
profiles/
profile_templates/
droid.properties
log4j2.properties
tmp/

### DIFF
--- a/cli-tests/test/cli-test.bats
+++ b/cli-tests/test/cli-test.bats
@@ -21,6 +21,22 @@ setup() {
   assert_output --partial 'usage: droid [options]'
 }
 
+@test "clean removes folders from droid home" {
+  $DROID_BIN --clean
+  run ls ~/.droid6
+  assert_output -p 'export_templates'
+  assert_output -p 'filter_definitions'
+  assert_output -p 'logs'
+  assert_output -p 'report_definitions'
+  refute_output -p 'container_sigs'
+  refute_output -p 'signature_files'
+  refute_output -p 'profiles'
+  refute_output -p 'profile_templates'
+  refute_output -p 'droid.properties'
+  refute_output -p 'log4j2.properties'
+  refute_output -p 'tmp'
+}
+
 @test "list the reports" {
   run $DROID_BIN -l
   assert_output -p "Report:	'Total unreadable folders'"
--- a/droid-command-line/src/main/java/uk/gov/nationalarchives/droid/command/action/CleanCommand.java
+++ b/droid-command-line/src/main/java/uk/gov/nationalarchives/droid/command/action/CleanCommand.java
@@ -1,3 +1,34 @@
+/*
+ * Copyright (c) 2016, The National Archives <pronom@nationalarchives.gov.uk>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following
+ * conditions are met:
+ *
+ *  * Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *
+ *  * Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ *  * Neither the name of the The National Archives nor the
+ *    names of its contributors may be used to endorse or promote products
+ *    derived from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+ * PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR
+ * CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+ * EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+ * PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+ * PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+ * LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
 package uk.gov.nationalarchives.droid.command.action;
 
 import org.apache.commons.io.FileUtils;
--- a/droid-command-line/src/main/java/uk/gov/nationalarchives/droid/command/action/CleanCommand.java
+++ b/droid-command-line/src/main/java/uk/gov/nationalarchives/droid/command/action/CleanCommand.java
@@ -1,0 +1,34 @@
+package uk.gov.nationalarchives.droid.command.action;
+
+import org.apache.commons.io.FileUtils;
+
+import java.nio.file.Path;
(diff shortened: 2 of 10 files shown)


cli-tests/test/cli-test.bats
```diff
@@ -21,6 +21,22 @@ setup() {
   assert_output --partial 'usage: droid [options]'
 }
 
+@test "clean removes folders from droid home" {
```
techncl, line 24: Do you think that it's worth asserting, before the `--clean`, that these folders exist?

droid-command-line/src/main/java/uk/gov/nationalarchives/droid/command/action/CleanCommand.java
```diff
@@ -0,0 +1,65 @@
...
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package uk.gov.nationalarchives.droid.command.action;
+
+import org.apache.commons.io.FileUtils;
+
+import java.nio.file.Path;
+import java.util.stream.Stream;
+
+public class CleanCommand implements DroidCommand {
```
techncl, line 39: 1. This name is a bit vague should we call it something like `CleanDroidHomeFoldersCommand`, `CleanDroid6FoldersCommand` or something?
2. Should we add a comment explaining why this command is deleting these folders or why this command is necessary?

droid-command-line/src/main/java/uk/gov/nationalarchives/droid/command/action/CommandLineParam.java
```diff
@@ -63,6 +63,14 @@ public DroidCommand getCommand(CommandFactory commandFactory, CommandLine cli) {
         }
     },
 
+    /** Clean. */
+    CLEAN("C", "clean", I18N.CLEAN_HELP) {
```
techncl, line 67: 1. Maybe `CLEAN_HOME_DIR("C", "clean-home", I18N.CLEAN_HELP)`?
2. Why not a lowercase `c`?

droid-command-line/src/test/java/uk/gov/nationalarchives/droid/command/action/CleanCommandTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @Before
+    public void setUp() throws IOException {
+        droidHome = Files.createTempDirectory("clean");
+    }
+
+    @Test
+    public void testCleanDeletesDirectories() throws CommandExecutionException, IOException {
+        Stream.of("container_sigs", "signature_files", "profiles", "profile_templates", "tmp", "not-included-directory").forEach(this::createDirectory);
+        Stream.of("droid.properties", "log4j2.properties", "not-included-file").forEach(this::createFile);
```
techncl, line 56: 1. Should we assert that the files/folders have been created beforehand?
2. do you think it's worth adding similar files e.g. "Profiles" (capitalised), "profiles2" in order to ensure that they don't get deleted?

droid-command-line/src/main/java/uk/gov/nationalarchives/droid/command/action/CleanCommand.java
```diff
@@ -0,0 +1,65 @@
...
+                "profiles",
+                "profile_templates",
+                "droid.properties",
+                "log4j2.properties",
+                "tmp"
+        );
+    }
+
+    @Override
+    public void execute() throws CommandExecutionException {
```
techncl, line 58: Should this maybe return a confirmation message (that tells them that files are deleted and potentially shows them what's in the directory now) or is the fact that it didn't throw an error, enough?

droid-command-line/src/main/java/uk/gov/nationalarchives/droid/command/action/CleanCommand.java
```diff
@@ -0,0 +1,65 @@
...
+
+import org.apache.commons.io.FileUtils;
+
+import java.nio.file.Path;
+import java.util.stream.Stream;
+
+public class CleanCommand implements DroidCommand {
+
+    private final Path droidHome;
+    private final Stream<String> toDelete;
```
techncl, line 42: `foldersToDelete` Maybe?

droid-command-line/src/main/java/uk/gov/nationalarchives/droid/command/action/CommandFactory.java
```diff
@@ -134,4 +134,5 @@ public interface CommandFactory {
     DroidCommand getListReportCommand();
 
 
+    DroidCommand getCleanCommand();
```
techncl, line 137: This is missing a doc comment above, like the others.

droid-command-line/src/test/java/uk/gov/nationalarchives/droid/command/action/CleanCommandTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+    private Path droidHome;
+
+    @Before
+    public void setUp() throws IOException {
+        droidHome = Files.createTempDirectory("clean");
+    }
+
+    @Test
+    public void testCleanDeletesDirectories() throws CommandExecutionException, IOException {
```
techncl, line 54: Maybe `testCleanDeletesDirectoriesAndFiles`?

techncl (commented):
Looks good! Just a few minor comments.

sparkhi:
> This will clean the folders out of the ~/.droid6 directory.
> 
> This is often what is recommended when DROID misbehaves so having a command to do this will be useful.
> 
> It will delete the following directories and files
> 
> container_sigs/ signature_files/ profiles/ profile_templates/ droid.properties log4j2.properties tmp/

I suggest excluding `profile_templates` from the cleanup. We do not supply any templates, so if there is anything in there, that is created by the users, so we should leave it as it is.

sparkhi:
> > This will clean the folders out of the ~/.droid6 directory.
> > This is often what is recommended when DROID misbehaves so having a command to do this will be useful.
> > It will delete the following directories and files
> > container_sigs/ signature_files/ profiles/ profile_templates/ droid.properties log4j2.properties tmp/
> 
> I suggest excluding `profile_templates` from the cleanup. We do not supply any templates, so if there is anything in there, that is created by the users, so we should leave it as it is.

Sorry, my mistake. I was thinking of `export_templates` folder. Please ignore the above comment

droid-command-line/src/main/java/uk/gov/nationalarchives/droid/command/action/CleanCommand.java
MancunianSam: 1. I don't think so. The command is called clean and it matches other classes like HelpCommand
2. There is help text which describes what it does and it's pretty clear from the code what it does. Clean commands are also fairly common. I'd rather not clutter it up with too many comments, there are enough as it is.

droid-command-line/src/main/java/uk/gov/nationalarchives/droid/command/action/CommandLineParam.java
MancunianSam: 1. The command is called "clean" which is the second argument and the enum entry CLEAN matches it.
2. It's already used in this file.

droid-command-line/src/main/java/uk/gov/nationalarchives/droid/command/action/CleanCommand.java
MancunianSam: This overrides a method from the parent `DroidCommand` class so it has to be `void execute()` so we can't return anything.

droid-command-line/src/main/java/uk/gov/nationalarchives/droid/command/action/CleanCommand.java
MancunianSam: They're not all folders, droid.properties, and log4j2.properties are files.

droid-command-line/src/test/java/uk/gov/nationalarchives/droid/command/action/CleanCommandTest.java
techncl: Ah, ok, I'm not too familiar with this part of DROID

cli-tests/test/cli-test.bats
MancunianSam: I don't think it matters. It makes the test quite a bit longer and all we really care about is whether they're there afterwards or not.

techncl (approved):
👍

PR #1341, opened 2025-05-20
Add checksum option to DROID API

MancunianSam:
This allows you to pass in a list of hash algorithms and the API will
return the checksums for the file for that algorithm.

### DIFF
--- a/droid-api/src/main/java/uk/gov/nationalarchives/droid/internal/api/DroidAPI.java
+++ b/droid-api/src/main/java/uk/gov/nationalarchives/droid/internal/api/DroidAPI.java
@@ -31,34 +31,42 @@
  */
 package uk.gov.nationalarchives.droid.internal.api;
 
+import java.io.*;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
 import java.nio.file.Files;
-import java.io.IOException;
 import java.nio.file.Path;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Optional;
-import java.util.ResourceBundle;
+import java.util.*;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.BiFunction;
 import java.util.stream.Collectors;
 
 import org.apache.commons.lang3.StringUtils;
 
 import org.apache.http.client.utils.URIBuilder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import software.amazon.awssdk.core.ResponseInputStream;
 import software.amazon.awssdk.core.exception.SdkClientException;
 import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.regions.providers.DefaultAwsRegionProviderChain;
 import software.amazon.awssdk.services.s3.S3Client;
 import software.amazon.awssdk.services.s3.S3Uri;
 import software.amazon.awssdk.services.s3.S3Utilities;
+import software.amazon.awssdk.services.s3.model.GetObjectRequest;
+import software.amazon.awssdk.services.s3.model.GetObjectResponse;
 import software.amazon.awssdk.services.s3.model.S3Object;
 import uk.gov.nationalarchives.droid.core.BinarySignatureIdentifier;
 import uk.gov.nationalarchives.droid.core.SignatureParseException;
 import uk.gov.nationalarchives.droid.core.interfaces.*;
 import uk.gov.nationalarchives.droid.core.interfaces.archive.ContainerIdentifier;
+import uk.gov.nationalarchives.droid.core.interfaces.hash.MD5HashGenerator;
+import uk.gov.nationalarchives.droid.core.interfaces.hash.SHA1HashGenerator;
+import uk.gov.nationalarchives.droid.core.interfaces.hash.SHA256HashGenerator;
+import uk.gov.nationalarchives.droid.core.interfaces.hash.SHA512HashGenerator;
 import uk.gov.nationalarchives.droid.core.interfaces.resource.*;
 
 
@@ -107,8 +115,21 @@ public final class DroidAPI implements AutoCloseable {
 
     private final HttpClient httpClient;
 
+    private final List<HashAlgorithm> hashAlgorithms;
 
-    private DroidAPI(DroidCore droidCore, ContainerIdentifier zipIdentifier, ContainerIdentifier ole2Identifier, ContainerIdentifier gzIdentifier, String containerSignatureVersion, String binarySignatureVersion, String droidVersion, S3Client s3Client, HttpClient httpClient, Region s3Region) {
+    private DroidAPI(
+            DroidCore droidCore,
+            ContainerIdentifier zipIdentifier,
+            ContainerIdentifier ole2Identifier,
+            ContainerIdentifier gzIdentifier,
+            String containerSignatureVersion,
+            String binarySignatureVersion,
+            String droidVersion,
+            S3Client s3Client,
+            HttpClient httpClient,
+            Region s3Region,
+            List<HashAlgorithm> hashAlgorithms
+    ) {
         this.droidCore = droidCore;
         this.zipIdentifier = zipIdentifier;
         this.ole2Identifier = ole2Identifier;
@@ -119,6 +140,7 @@ public final class DroidAPI implements AutoCloseable {
         this.s3Region = getRegionOrDefault(s3Region);
         this.s3Client = getS3ClientOrDefault(s3Client);
         this.httpClient = getHttpClientOrDefault(httpClient);
+        this.hashAlgorithms = hashAlgorithms;
     }
 
     private HttpClient getHttpClientOrDefault(HttpClient httpClient) {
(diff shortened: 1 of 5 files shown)


droid-api/src/main/java/uk/gov/nationalarchives/droid/internal/api/DroidAPI.java
```diff
@@ -243,11 +269,61 @@ private List<ApiResult> submitHttpIdentification(final URI uri, String extension
...
         try (final HttpIdentificationRequest request = new HttpIdentificationRequest(metaData, id, httpClient)) {
             request.setExtension(extension);
             request.open(uri);
-            return getApiResults(request);
+            return getApiResults(request, hashResults);
+        }
+    }
+
+    private <T> Map<HashAlgorithm, String> generateHashResults(T identifier, BiFunction<HashAlgorithm, T, String> hashFunction) {
+        return hashAlgorithms.stream().collect(Collectors.toMap(
```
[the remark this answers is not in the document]
MancunianSam: Ah good spot. And the test was always passing in an empty list so I didn't notice. The builder will now return an empty list if hasAlgorithms isn't set and I've updated the test to pass null in to check that it works.

droid-api/src/main/java/uk/gov/nationalarchives/droid/internal/api/DroidAPI.java
```diff
@@ -192,7 +223,8 @@ public DroidAPI build() throws SignatureParseException {
             String containerVersion = StringUtils.substringAfterLast(containerSignature.getFileName().toString(), "-").split("\\.")[0];
             String droidVersion = ResourceBundle.getBundle("options").getString("version_no");
             ContainerApi containerApi = new ContainerApi(droidCore, containerSignature);
-            return new DroidAPI(droidCore, containerApi.zipIdentifier(), containerApi.ole2Identifier(), containerApi.gzIdentifier(), containerVersion, droidCore.getSigFile().getVersion(), droidVersion, this.s3Client, this.httpClient, this.s3Region);
+            List<HashAlgorithm> hashAlgorithmsOrEmptyList = this.hashAlgorithms == null ? Collections.emptyList() : this.hashAlgorithms;
```
techncl, line 226: This might be my ignorance of Java, but on line 182, couldn't you just do

`private List<HashAlgorithm> hashAlgorithms = Collections.emptyList();` or 
`private List<HashAlgorithm> hashAlgorithmsOrEmptyList = Collections.emptyList();`?

MancunianSam: I think you're right, I've changed the builder so it calls `private List<HashAlgorithm> hashAlgorithms = Collections.emptyList();`

droid-api/src/main/java/uk/gov/nationalarchives/droid/internal/api/DroidAPI.java
```diff
@@ -243,18 +274,68 @@ private List<ApiResult> submitHttpIdentification(final URI uri, String extension
...
 
-    private List<ApiResult> submitS3Identification(final URI uri, String extension) throws IOException {
+    private <T> Map<HashAlgorithm, String> generateHashResults(T identifier, BiFunction<HashAlgorithm, T, String> hashFunction) {
+        return hashAlgorithms.stream().collect(Collectors.toMap(
+                algorithm -> algorithm,
+                algorithm -> hashFunction.apply(algorithm, identifier)
+        ));
+    }
+
+    private String getFileHash(HashAlgorithm hashAlgorithm, Path path) {
```
sparkhi, line 293: Is the FileInputStream staying open here? 
Might be easier to use try-with-resource so it closes automatically 

```
        try (InputStream fs = new FileInputStream(path.toFile())){
            return getHash(hashAlgorithm, fs);
        } catch (IOException e) {
            throw new RuntimeException(e);
        }
```

droid-api/src/main/java/uk/gov/nationalarchives/droid/internal/api/DroidAPI.java
```diff
@@ -243,18 +274,68 @@ private List<ApiResult> submitHttpIdentification(final URI uri, String extension
...
+            return getHash(hashAlgorithm, new FileInputStream(path.toFile()));
+        } catch (FileNotFoundException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private String getS3Hash(HashAlgorithm algorithm, S3Uri s3Uri) {
+        String key = s3Uri.key().orElseThrow(() -> new RuntimeException("Key not found in uri " + s3Uri.uri()));
+        String bucket = s3Uri.bucket().orElseThrow(() -> new RuntimeException("Bucket not found in uri " + s3Uri.uri()));
+        ResponseInputStream<GetObjectResponse> responseInputStream = s3Client.getObject(GetObjectRequest.builder().bucket(bucket).key(key).build());
```
sparkhi, line 304: similar to above, try-with-resource to close the `ResponseInputStream`

sparkhi (commented):
A couple of comments. Looks good. Do we need to convey the `ApiResult` vs `APIResult` breaking change to TDR in advance. 

... and may I tempt you to update this page too -> https://github.com/digital-preservation/droid/wiki/How-to-use-DROID-internal-API  (sorry!!)

sparkhi:
> A couple of comments inline. Looks good. Do we need to convey the `ApiResult` vs `APIResult` breaking change to TDR in advance.
> 
> ... and may I tempt you to update this page too -> https://github.com/digital-preservation/droid/wiki/How-to-use-DROID-internal-API (sorry!!)

I had another look at the "how to" page and it has been out of date for a while, feel free to ignore it for this PR.

droid-api/src/main/java/uk/gov/nationalarchives/droid/internal/api/DroidAPI.java
sparkhi: You would still need the null protection in case someone directly uses `Builder` and passes in a null. Might be easier to null check in `public DroidAPIBuilder hashAlgorithms(final List<HashAlgorithm> hashAlgorithms)`

MancunianSam: I did think that but then I thought, if you're explicitly passing null into a builder then you deserve what happens to you. That being said, I've added a null check in the builder method.

droid-api/src/main/java/uk/gov/nationalarchives/droid/internal/api/DroidAPI.java
```diff
@@ -322,17 +424,17 @@ private <T> List<ApiResult> getApiResults(IdentificationRequest<T> request) thro
         boolean fileExtensionMismatch = resultCollection.getExtensionMismatch();
 
         return resultCollection.getResults()
-                .stream().map(res -> createApiResult(res, extension, fileExtensionMismatch, request.getIdentifier().getUri()))
+                .stream().map(res -> createIdentificationResult(res, extension, fileExtensionMismatch, request.getIdentifier().getUri()))
                 .collect(Collectors.toList());
     }
 
-    private ApiResult createApiResult(IdentificationResult result, String extension, boolean extensionMismatch, URI uri) {
+    private IdentificationResult createIdentificationResult(uk.gov.nationalarchives.droid.core.interfaces.IdentificationResult result, String extension, boolean extensionMismatch, URI uri) {
```
techncl, line 431: Can't you just use IdentificationResult or interfaces.IdentificationResult?

droid-api/src/main/java/uk/gov/nationalarchives/droid/internal/api/DroidAPI.java
```diff
@@ -344,7 +446,7 @@ private <T> IdentificationResultCollection identifyByExtension(final Identificat
     private Optional<String> getContainerPuid(final IdentificationResultCollection binaryResult) {
         List<String> containerPuids = Arrays.asList(ZIP_PUID, OLE2_PUID, GZIP_PUID);
         return binaryResult.getResults().stream()
-                .map(IdentificationResult::getPuid)
+                .map(uk.gov.nationalarchives.droid.core.interfaces.IdentificationResult::getPuid)
```
techncl, line 449: Can't you just use IdentificationResult or interfaces.IdentificationResult?

droid-api/src/main/java/uk/gov/nationalarchives/droid/internal/api/DroidAPI.java
MancunianSam: Good point, I've changed it.

droid-api/src/main/java/uk/gov/nationalarchives/droid/internal/api/DroidAPI.java
MancunianSam: I've changed this one as well.

droid-api/src/main/java/uk/gov/nationalarchives/droid/internal/api/DroidAPI.java
MancunianSam: Actually, I've changed my mind, I've renamed the record APIIdentificationResult so this doesn't need the package qualifier now.

droid-api/src/main/java/uk/gov/nationalarchives/droid/internal/api/HashAlgorithm.java
```diff
@@ -31,48 +31,13 @@
...
-
-    public boolean isFileExtensionMismatch() {
-        return fileExtensionMismatch;
-    }
-
-    public URI getUri() {
-        return uri;
-    }
+public enum HashAlgorithm {
+    /** MD5. **/
```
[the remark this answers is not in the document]
MancunianSam: Technically they are because of the checkstyle rules but I agree that this is daft so I've disabled that rule for this class.

sparkhi (approved):

PR #1347, opened 2025-05-29
DR2-2241 DROID help pages.

MancunianSam:
This has replaced the html pages which were used to generate the swing
FAQ.


sparkhi (approved):
Looks good to me.

PR #1360, opened 2025-06-04
Fix S3 identification failure on OLE2 containers.

MancunianSam:
For OLE2 parsing we use Apache POI.

This works out the maximum length of the file by multiplying the number of sectors in the file by the size of the sectors.
Their own code comments say that for a sector size of 512 (which .doc is) they may over-estimate the size of the file by 65Kb. So DROID was
asking S3 for a byte range which was larger than the size of the file and S3 was understandably returning an error.

I don't understand the internals completely so I don't know how many .doc files are affected but it's definitely some of them.

I've added in a check so that if DROID tries to access a range that is larger than the file, it will return null. This is working with the
failing test file.


sparkhi (approved):
Looks fine to me, although I did not fully understand the reason, I was wondering whether the null return should be only when it exceeds the length rather than on equal as well?

MancunianSam:
It's uses for a byte range and the bytes are zero indexed so the last byte is `length - 1`

PR #1363, opened 2025-06-04
Fix PR creation call.

MancunianSam:
Because of the single quotes, the branch name is literally $BRANCH_NAME
which isn't valid.

This uses the actual value of the environment variable.


techncl (approved):
👍

PR #1441, opened 2025-09-15
Use larger window size.

MancunianSam:
4K is far to small for a window size when you're making http calls.

I've increased this to 4Mb and it seems to be working.


sparkhi (commented):
Just one suggestion

PR #1621, opened 2026-03-18
Load the ResourceSelectorDialog asynchronously

MancunianSam:
This takes up to 4 seconds on a Windows machine. If you load it
asynchronously, it should complete before the profile is loaded so users
should never see an incomplete resource dialog.


sparkhi (approved):
👍🏼

